PR #34, opened 2015-01-16
Standardized setup

matthuhiggins:
* Add a dummy app!
* Introduces logs, to see what the tests are doing!
* Remove manual setup of models and db!
* Allows running of index.rake tasks from the project root!
* Introduces new way to define the percolator index up front, so that it's not created on demand!

@ebarendt @alanpearson


lib/elastic_record/index.rb
```diff
@@ -81,8 +82,10 @@ def get(end_path, json = nil)
     end
 
     private
+
       def new_index_name
         "#{alias_name}_#{(Time.now.to_f * 100).to_i}"
```
ebarendt: Can we use `yyyymmdd` instead of `to_f`?

lib/elastic_record/index/percolator.rb
```diff
@@ -2,9 +2,9 @@ module ElasticRecord
   class Index
     module Percolator
       def create_percolator(name, elastic_query)
-        unless exists? percolator_name
-          create percolator_name
-        end
+        # unless exists? percolator_name
```
ebarendt: Need to create the percolator first?

test/helper.rb
```diff
@@ -1,14 +1,8 @@
-require 'bundler/setup'
-Bundler.require
+ENV["RAILS_ENV"] = "test"
```
ebarendt: Isn't this automagic?

matthuhiggins:
No reason to create the percolator like that anymore. It will be created with `rake index:create` now, as long as your model has:

```ruby
class Product
  elastic_index.has_percolator = true
end
```

lib/elastic_record/index.rb
matthuhiggins: I would probably go with dashes, "yyyy-mm-dd"

PR #35, opened 2015-03-06
Use Oj instead of ActiveSupport::JSON

malept:
Testing a script that creates 250,000 records (in bulk batches of 500) showed a significant speed increase (~62%).

### DIFF
--- a/elastic_record.gemspec
+++ b/elastic_record.gemspec
@@ -20,4 +20,5 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'arelastic', '>= 0.7.0'
   s.add_dependency 'activemodel'
+  s.add_dependency 'oj'
 end
--- a/lib/elastic_record/connection.rb
+++ b/lib/elastic_record/connection.rb
@@ -1,4 +1,5 @@
 require 'net/http'
+require 'oj'
 
 module ElasticRecord
   class Connection
@@ -37,10 +38,10 @@ module ElasticRecord
     end
 
     def json_request(method, path, json)
-      body = json.is_a?(Hash) ? ActiveSupport::JSON.encode(json) : json
+      body = json.is_a?(Hash) ? Oj.dump(json, :mode => :compat) : json
       response = http_request_with_retry(method, path, body)
 
-      json = ActiveSupport::JSON.decode response.body
+      json = Oj.compat_load(response.body)
       raise ConnectionError.new(response.code, json['error']) if json['error']
 
       json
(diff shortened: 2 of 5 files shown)


dcromer:
:+1:

dcromer:
From http://www.ohler.com/oj/
```
Oj.default_options = {:mode => :compat }
```

lib/elastic_record/connection.rb
```diff
@@ -37,10 +38,10 @@ def json_delete(path, json = nil)
     end
 
     def json_request(method, path, json)
-      body = json.is_a?(Hash) ? ActiveSupport::JSON.encode(json) : json
+      body = json.is_a?(Hash) ? Oj.dump(json, :mode => :compat) : json
       response = http_request_with_retry(method, path, body)
 
-      json = ActiveSupport::JSON.decode response.body
+      json = Oj.compat_load(response.body)
```
dcromer: Should we be using `strict_load` here?

malept: It breaks tests if I use `load` or `strict_load`.

malept:
I specifically used `Oj.dump(json, :mode => :compat)` instead of `Oj.default_options` so that `elastic_record`'s options don't override options within the rest of the ER-using app.

lib/elastic_record/connection.rb
dcromer: OK.  I ask because I know that Oj can demarshall arbitrary objects if it's not configured properly, as the subscription bot was doing.  I can't figure out if compat mode is susceptible to this from the documentation.

dcromer:
Ahh, that makes sense.

matthuhiggins:
Does this produce the exact same output?

malept:
> Does this produce the exact same output?

From what I've read about Oj's `:compat` modes, yes.

matthuhiggins:
I would rather verify with raw data.

malept:
I did some verification with raw data. There are two classes of differences that I have seen:

`ActiveSupport::JSON` | Oj Gem
--------------------- | ------
Unicode-encodes (`\u1234`) HTML special characters (e.g., `&<>`) | Does not encode HTML special chars (unless you tell it to - in which case it will backslash-escape them)
Does not round certain floats (e.g., 12.345678999999996) | Rounds certain floats (e.g., 12.345679)

matthuhiggins:
Seems all good to me.

elastic_record.gemspec
```diff
@@ -20,4 +20,5 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'arelastic', '>= 0.7.0'
   s.add_dependency 'activemodel'
+  s.add_dependency 'oj'
```
matthuhiggins: I would like to see this removed. Something like `ElasticRecord.json_parser = :oj`.

PR #36, opened 2015-03-10
Use Oj optionally

malept:
This reverts the default JSON parser to `ActiveSupport::JSON`, and only uses Oj if:

1. `ElasticRecord.json_parser` is set to `:oj`
2. Oj is loaded

This should ease upgrading.

### DIFF
--- a/lib/elastic_record.rb
+++ b/lib/elastic_record.rb
@@ -17,6 +17,14 @@ module ElasticRecord
     def configure
       yield(ElasticRecord::Config)
     end
+
+    def json_parser
+      @@json_parser ||= :active_support
+    end
+
+    def json_parser=(value)
+      @@json_parser = value
+    end
   end
 end
 
--- a/lib/elastic_record/connection.rb
+++ b/lib/elastic_record/connection.rb
@@ -48,7 +48,7 @@ module ElasticRecord
 
     def json_encode(data)
       if ElasticRecord.json_parser == :oj
-        Oj.dump(data, :mode => :compat)
+        Oj.dump(data, mode: :compat)
       else
         ActiveSupport::JSON.encode(data)
       end
--- a/lib/elastic_record/connection.rb
+++ b/lib/elastic_record/connection.rb
@@ -1,5 +1,4 @@
 require 'net/http'
-require 'oj'
 
 module ElasticRecord
   class Connection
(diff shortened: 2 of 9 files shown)


lib/elastic_record/connection.rb
```diff
@@ -38,15 +37,31 @@ def json_delete(path, json = nil)
...
-      json = Oj.compat_load(response.body)
+      json = json_decode(response.body)
       raise ConnectionError.new(response.code, json['error']) if json['error']
 
       json
     end
 
+    def json_encode(data)
+      if ElasticRecord.json_parser == :oj
+        Oj.dump(data, :mode => :compat)
```
matthuhiggins: ruby 1.9 hash syntax

malept: Fixed in a47c87f.

lib/elastic_record/index/deferred.rb
```diff
@@ -42,8 +42,10 @@ def flush!
 
         private
           READ_METHODS = [:json_get, :head]
+          SERIALIZE_METHODS = [:json_encode, :json_decode]
           def method_missing(method, *args, &block)
             super unless index.real_connection.respond_to?(method)
+            return index.real_connection.send(method, *args, &block) if SERIALIZE_METHODS.include?(method)
```
matthuhiggins: I don't think that json_encode/json_decode should be defined on connection.rb.

lib/elastic_record/index/documents.rb
```diff
@@ -64,7 +64,7 @@ def bulk(options = {})
         yield
 
         if current_bulk_batch.any?
-          body = current_bulk_batch.map { |action| "#{Oj.dump(action, :mode => :compat)}\n" }.join
+          body = current_bulk_batch.map { |action| "#{connection.json_encode(action)}\n" }.join
```
matthuhiggins: Thoughts on `ElasticRecord::JSON.json_encode`?

lib/elastic_record.rb
```diff
@@ -17,6 +17,14 @@ class << self
...
       yield(ElasticRecord::Config)
     end
+
+    def json_parser
+      @@json_parser ||= :active_support
+    end
+
+    def json_parser=(value)
+      @@json_parser = value
+    end
```
matthuhiggins: Thoughts on Thoughts on `ElasticRecord::JSON.parser = :oj`?

malept:
> Thoughts on `ElasticRecord::JSON.encode`?
> Thoughts on `ElasticRecord::JSON.parser = :oj`?

The refactor suggestions are fine, but at this point, I'm wondering where the line is to stop reinventing the wheel and switch to `multi_json`.

matthuhiggins:
I wish that could be an obvious suggestion, but have you read https://github.com/rails/rails/pull/10576?

malept:
I had not, until now. I was not aware of how much overhead it is, though I shouldn't be surprised.

malept:
Refactored in ce9370f.

PR #43, opened 2015-12-11
Make testsuite more consistent

malept:
It was hard to see the actual test failures due to the state of the testsuite. This makes sure all of the relational databases/tables are created correctly, and that some of the more obviously intermittent errors in the testcases are fixed.


test/elastic_record/integration/active_record_test.rb
```diff
@@ -48,7 +48,7 @@ def setup
     setup_project_database(
       'adapter'   => 'mysql2',
       'host'      => "localhost",
-      'database'  => 'elastic_record_test',
```
matthuhiggins: why subtest?

test/elastic_record/relation_test.rb
```diff
@@ -28,12 +28,14 @@ def test_to_hits
...
 
   def test_to_ids
+    Widget.elastic_index.delete_all
     create_widgets [Widget.new(id: 5, color: 'red'), Widget.new(id: 10, color: 'blue')]
 
     assert_equal ['5', '10'].to_set, Widget.elastic_relation.to_ids.to_set
   end
 
   def test_to_a
+    Widget.elastic_index.delete_all
```
matthuhiggins: :-1: This should not be necessary

matthuhiggins:
mostly good

test/elastic_record/integration/active_record_test.rb
malept: Seems like a bad idea to name it the same as the database you're using for the rest of the tests, if you're dropping/creating it within specific testcases.

test/elastic_record/relation_test.rb
malept: I agree, but these tests assume that the indexes are empty at the beginning, and in running this testsuite over and over and over, this is only the case sometimes.

test/elastic_record/integration/active_record_test.rb
matthuhiggins: Ah, this code runs after the test suite starts

test/elastic_record/relation_test.rb
matthuhiggins: Been running these tests and have not seen this test fail.

matthuhiggins: Hah, 5 minutes later, I saw this test fail.

malept: I am happy to implement a different solution to fix these intermittent failures.

matthuhiggins: We do not have these failures in our apps, which do more complex things than here. I believe that is enough evidence to show that something is going wrong here.

This gem was originally made without a test/dummy app. I am thinking of changing the test suite to use real active_record models found in test/dummy/app/models.

ebarendt: We currently have two randomly failing tests. One in content_system is skipped. I think the other one is also in content_system.

matthuhiggins: That's due to scoring and ordering, which is different than these failures, which are due to bad persistence state when the test starts.

malept: I propose to merge this PR so we can see (real) test failures better, then you can delete my hacks once you change the testsuite to use the test/dummy app pattern.

test/elastic_record/index/mapping_test.rb
```diff
@@ -2,7 +2,7 @@
 
 class ElasticRecord::Index::MappingTest < MiniTest::Test
   def test_delete_mapping
-    index_name = index.create
+    index_name = index.create("widgets_#{SecureRandom.random_number(100_000_000_000_000)}")
```
matthuhiggins: Hardcode a name that is not used elsewhere, rather than random.

malept: 5cba5487354f5e1b0513ddc7ce6753c4f7ff9bad

matthuhiggins:
Looks good @malept. I will do a comprehensive test suite update during my post-xmas/pre-new-years week,

PR #45, opened 2016-03-22
Update to ElasticSearch 2.2

matthuhiggins:
- [x] [Types may no longer be deleted](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#_types_may_no_longer_be_deleted)
- [x] [Delete by query is now a plugin](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_removed_features.html#_delete_by_query_is_now_a_plugin)


test/elastic_record/index/documents_test.rb
```diff
@@ -89,27 +101,28 @@ def test_bulk
...
+  #   index.bulk do
+  #     InheritedWidget.elastic_index.index_document '5', color: 'green'
+  #
+  #     expected = [
+  #       {index: {_index: index.alias_name, _type: "widget", _id: "5"}},
+  #       {color: "green"}
+  #     ]
+  #     assert_equal expected, index.current_bulk_batch
+  #   end
+  # end
```
malept: Why was this commented out instead of deleted? Will the tests be fixed?

matthuhiggins: I'm working on it. It's really tricky with the deferred system, since this test is trying to test something about the connection, but we have a fake connection.

This test is also a problem in master. [This causes trouble](https://github.com/data-axle/elastic_record/blob/master/test/elastic_record/index/documents_test.rb#L10-L14).

malept: Oh, OK. I had assumed that the PR was merge-ready. I'll wait to look at the changes further after you give the all-clear sign.

.travis.yml
```diff
@@ -1,6 +1,12 @@
 rvm: 2.2.3
 sudo: false
 cache: bundler
+addons:
+  apt:
+    sources:
+    - elasticsearch-2.1
+    packages:
+    - elasticsearch
```
matthuhiggins: Thanks! I'm targeting 2.2 though.

malept: I know, but Travis doesn't have sources for ES 2.2 yet :disappointed:

matthuhiggins: aah man. Well both should work.

malept: The reason why I moved the Travis config to Trusty was because the elasticsearch-2.1 source doesn't actually work. They are also not adding any more whitelisted sources, so the only way to test against ES 2.x is to either use Elastic.co's package repository (which only provides the latest 2.x package), or install a specific package from their archive download.

malept: I verified that ES 1.4.0 was being run despite setting `elasticsearch-2.1` as the source by running `curl localhost:9200` in `before_script`, if you would like to check my work.

matthuhiggins: aaah crap

PR #47, opened 2016-05-24
Use as_search if it's available, even if our index mapping is not an

dcromer:
object.

### DIFF
--- a/lib/elastic_record/callbacks.rb
+++ b/lib/elastic_record/callbacks.rb
@@ -60,7 +60,11 @@ module ElasticRecord
         when :nested
           value.map(&:as_search)
         else
-          value
+          if value.respond_to?(:as_search)
+            value.as_search
+          else
+            value
+          end
         end
 
       if value.present? || value == false


lib/elastic_record/callbacks.rb
```diff
@@ -60,7 +60,11 @@ def elastic_search_value(field, mapping)
         when :nested
           value.map(&:as_search)
         else
```
matthuhiggins: There ought to be a case statement for `: geo_point`.

dcromer: So, your attribute definition looks like:

```
  "manual_geocoordinate": {
    "type": "geo_point",
    "class_name": "GeoPoint",
    "attributes": {
      "lat": {
        "type": "float"
      },
      "lon": {
        "type": "float"
      }
    }
  },
```

And then [axle_attributes](https://github.com/data-axle/axle_attributes/blob/4c130518b993d40c6de7ea79dfa5b988e7ba80d6/lib/axle_attributes/definition.rb) would translate that into elastic_record/superstore-speak?

matthuhiggins: I am starting to like your geopoint type that is native in superstore.

ebarendt:
Is this still needed?

PR #51, opened 2017-08-21
PercolatorModel represents registered percolator queries

dcromer:
# Motivation

Percolators have changed significantly since ElasticSearch 1.x, especially into 5.x.  "Percolator indexes" are more like normal indexes now, with a special mapping, instead of being a special kind of index.  Furthermore, storing all percolators for a given doctype in the same index for that doctype is often times an anti-pattern.  It's better to store them in a separate index.

This PR introduces a new model, `ElasticRecord::PercolatorModel`, similar to `ElasticRecord::Model`, that abstracts a model as a registered percolator query for a given doctype.  It completely replaces `ElasticRecord::Index::Percolator".

# Questions/TODO

- [x] - Remove ElasticRecord::Index::Percolator, `has_percolator`, and all percolator management features.
- [x] - Does the current ElasticRecord index lifecycle management make sense for percolator indexes?  Do percolator indexes need to be aliased?
- [x] - How should analyzers be managed?  Pulling the mapping for the target type off of the target model makes sense, I think, but the target model mapping also relies on certain analyzers being present.
    * Option:  merge all analyzers from the target model's elastic_index settings into the percolator model's elastic_index settings.
- [x] - Does it make sense to support partial document updates for registered queries?  I don't think it does.


lib/elastic_record/index.rb
```diff
@@ -84,14 +79,14 @@ def configure(&block)
     end
 
     def get(end_path, json = nil)
-      connection.json_get "/#{alias_name}/#{type}/#{end_path}", json
+      doctype = doctypes.first
```
dcromer: This is the only problem with doctypes right now:  sometimes, operations have to specify a doctype, so we need to know which one.

lib/elastic_record/index/configurator.rb
```diff
@@ -2,14 +2,11 @@ module ElasticRecord
...
       attr_reader :index
+
       def initialize(index)
         @index = index
       end
 
-      def property(name, options)
-        index.mapping[:properties][name.to_sym] = options
-      end
-
```
dcromer: Unused.

lib/elastic_record/index/deferred.rb
```diff
@@ -14,9 +14,8 @@ def run(index)
         attr_accessor :bulk_stack
 
         def initialize(index)
-          self.index = index
-          self.bulk_stack = []
-          @deferring_enabled = false
```
dcromer: unused.

lib/elastic_record/as_search.rb
```diff
@@ -0,0 +1,55 @@
...
+          end
+
+        if value.present? || value == false
+          value
+        end
+      rescue
+        raise "Field not found for #{field.inspect}"
+      end
+  end
+end
```
matthuhiggins: as_document.rb

lib/elastic_record/index.rb
```diff
@@ -75,6 +66,7 @@ def enable!
       @disabled = false
     end
 
+    # This is still kind of silly
```
matthuhiggins: Yeah... I think we can use the new ?refresh option and get rid of deferred.

lib/elastic_record/percolator_model.rb
```diff
@@ -0,0 +1,78 @@
...
+        extend ClassMethods
+        include InstanceMethods
+      end
+    end
+
+    module ClassMethods
+      def elastic_index
+        @elastic_index ||=
+          begin
+            index = ElasticRecord::Index.new([self, target_model])
```
matthuhiggins: Can be inlined

test/dummy/app/models/mock_model.rb
```diff
@@ -0,0 +1,108 @@
...
+    @persisted
+  end
+
+  def destroyed?
+    @destroyed
+  end
+
+  def reload
+  end
+end
```
matthuhiggins: Now that there's a dummy app, I've been trying to use real models. Until then, there's this!

test/elastic_record/index/configurator_test.rb
```diff
@@ -1,12 +1,7 @@
 require 'helper'
 
 class ElasticRecord::Index::ConfiguratorTest < MiniTest::Test
-  def test_property
-    configurator.property :am_i_cool, type: "boolean"
-
-    expected = {type: "boolean"}
-    assert_equal expected, configurator.index.mapping[:properties][:am_i_cool]
-  end
+  # TODO:  will probably end up going away entirely
```
matthuhiggins: I never got to fleshing this out. We don't use it.

test/elastic_record/relation/admin_test.rb
```diff
@@ -1,19 +1,12 @@
...
 class ElasticRecord::Relation::AdminTest < MiniTest::Test
-  def test_create_percolator
-    index.delete_percolator('green') if index.percolator_exists?('green')
-
-    relation = Widget.elastic_relation.filter('color' => 'green')
-    relation.create_percolator('green')
-
-    assert_equal relation.as_elastic, index.get_percolator('green')
+  def test_create_warmer
   end
```
matthuhiggins: Bad merge. Warmer is dead.

lib/elastic_record/index.rb
dcromer: I'm not following.   I was referring to how the elastic_connection is a property of the model, and the index has to reach into the model to get to it.

dcromer:
> Yes, I think percolators should be aliased, for all the same reason a document index should be aliased. No required here, but if easy, go for it.

Ok, we get that for free since `PercolatorModel` extends `Model`

lib/elastic_record/index.rb
matthuhiggins: Oh, thought you meant deferred is silly.

lib/elastic_record/percolator_model.rb
```diff
@@ -0,0 +1,51 @@
...
+            index.partial_updates = false
+            index.settings = target_model.elastic_index.settings.slice('analysis')
```
dcromer: This is kind of gross.  What do you think about moving analyzer configuration into the doctype, and then having the index build its `settings.analysis` from the registered doctypes, like I'm doing with mapping now?

README.md
```diff
@@ -153,23 +153,40 @@ end
...
+as an ElasticSearch query.
 
 ```ruby
-class Product < ActiveRecord::Base
-  elastic_index.configure do
-    property :status, type: "keyword"
+class ProductQuery
+  include ElasticRecord::PercolatorModel
+
+  target_model Product
```
matthuhiggins: Generic. Some ideas:
```ruby
percolates_model Product
percolates_on Product
```

README.md
```diff
@@ -153,23 +153,40 @@ end
...
+    Product.filter(status: status).as_elastic
   end
 end
 ```
 
-You can also directly access Product.elastic_index.mapping and Product.elastic_index.settings:
+Then, target models can be percolated.
+
+```
+  matching_queries = ProductQuery.percolate(product)
```
matthuhiggins: What is returned here? `Product` instances?

lib/elastic_record/as_document.rb
```diff
@@ -0,0 +1,50 @@
...
+      doctype.mapping[:properties].each do |field, mapping|
+        value = elastic_search_value field, mapping
+
+        unless value.nil?
+          json[field] = value
+        end
+      end
+
+      json
+    end
```
matthuhiggins: Can use `each_with_object`. Original version of this was trying to avoid rails dependencies. The gem is now rails dependent, and each_with_object is in ruby.

lib/elastic_record/as_document.rb
```diff
@@ -0,0 +1,50 @@
...
+              when :nested
+                value.map(&:as_search_document)
+              else
+                value
+              end
+
+      if value.present? || value == false
+        value
+      end
+      end
```
matthuhiggins: ⬅️

lib/elastic_record/doctype.rb
```diff
@@ -0,0 +1,43 @@
...
+      mapping.deep_merge!(custom_mapping)
+    end
+
+    def analysis=(custom_analysis)
+      analysis.deep_merge!(custom_analysis)
+    end
+
+    def ==(other)
+      name == other.name && mapping == other.mapping && analysis == other.analysis
+    end
```
matthuhiggins: What is the reason for defining `==`?

lib/elastic_record/index.rb
```diff
@@ -25,30 +23,30 @@ module ElasticRecord
...
     attr_accessor :partial_updates
 
-    def initialize(model)
-      @model = model
+    def initialize(models)
+      models = Array.wrap(models)
+      @model = models.first
+      @doctypes = models.map(&:doctype)
       @disabled = false
     end
```
matthuhiggins: The support for multiple doctypes, alongside the arbitrary selection of the first model, is disgruntling. Is the plural support necessary?

lib/elastic_record/index.rb
```diff
@@ -79,19 +69,18 @@ def real_connection
       model.elastic_connection
     end
 
-    def configure(&block)
-      Configurator.new(self).instance_eval(&block)
-    end
+    def get(end_path, doctype = nil, json = nil)
+      path = "/#{alias_name}"
+      path += "/#{doctype.name}" if doctype
+      path += "/#{end_path}"
```
matthuhiggins:
```ruby
path = File.join [alias_name, doctype, endpath].compact
```

(requires db57b91612f2b74595fd11da07ff2b398b4d7773)

lib/elastic_record/as_document.rb
dcromer: Ok, great.  There's some places we can use ActiveSupport::Concern as well.

lib/elastic_record/doctype.rb
dcromer: tests

lib/elastic_record/index.rb
dcromer: The goal was to model query documents as regular documents.  This has the advantage of eliminating all of the special percolator-specific code.  Indexes with percolators now just have two doctypes, and the necessary analysis settings and mappings are built automatically.

From elasticsearch's perspective, multiple doctype support should actually [be expected!](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#mapping-type)  We are restricting ElasticSearch by omitting it.

But, yes, from the `ElasticRecord::Model`'s perspective, doctype specification is redundant.

README.md
dcromer: A `ProductQuery` relation.

lib/elastic_record/index.rb
dcromer: believe it or not, this causes test failures.  I can't figure out why.

dcromer:
@matthuhiggins I want to release this.

elastic_record.gemspec
```diff
@@ -2,7 +2,7 @@
 
 Gem::Specification.new do |s|
   s.name = 'elastic_record'
-  s.version = '4.0.0'
+  s.version = '5.0.0'
```
matthuhiggins: 4.1.0

lib/elastic_record/percolator_model.rb
```diff
@@ -0,0 +1,44 @@
...
+              "document"      => document
+            }
+          }
+        }
+
+        hits = elastic_index.search(query)['hits']['hits']
+        ids = hits.map { |hits| hits['_id'] }
+
+        where(id: ids)
+      end
```
matthuhiggins: This is my only request: This method should take a model record, and not a hash.

The raw document stuff can live in `elastic_index`, or a separate method. My primary interest is that the public API supports a `Product` instance. (given that it implements `as_search_document`)

dcromer: I originally did this, but when it came to actually consuming the API, it made the percolation logic more complicated than necessary.  Presenting a document for percolation is not necessarily the responsibility of the class that represents the index query (PercolatorModel).

matthuhiggins: Alright, I'll move on :)

matthuhiggins:
v4.1.0 is released to rubygems

PR #52, opened 2017-09-01
Add a lookup for the current index name.

ebarendt:
Cached in-memory for 60s


test/elastic_record/name_cache_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'helper'
+
+class ElasticRecord::NameCacheTest < MiniTest::Test
+  class Felon
+    include TestModel
+  end
+
+  def test_index_name
+    Felon.elastic_index.create_and_deploy("felons_datestamp")
```
matthuhiggins: This will leave behind an index on every test run.

PR #53, opened 2017-09-07
Add configuration to load documents from source

bullfight:
To fetch documents without an additional request to a backing ActiveRecord database you can load the documents from `_source`.

```ruby
Product.elastic_index.loading_from_source do
  Product.elastic_search.filter(name: "Pizza")
end

```

Use `elastic_index.load_from_source = true` to configure an index without ActiveRecord.

```ruby
class Product
  include ActiveModel::Model
  include ElasticRecord::Record

  self.elastic_index.load_from_source = true
end
```

### DIFF
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -66,16 +66,6 @@ module ElasticRecord
       @disabled = false
     end
 
-    def load_from_source(&block)
-      if block_given?
-        @load_from_source = true
-        yield
-        @load_from_source = false
-      else
-        @load_from_source
-      end
-    end
-
     def real_connection
       model.elastic_connection
     end
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -66,6 +66,13 @@ module ElasticRecord
       @disabled = false
     end
 
+    def loading_from_source(&block)
+      @load_from_source = true
+      yield
+    ensure
+      @load_from_source = false
+    end
+
     def real_connection
       model.elastic_connection
     end
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -37,6 +37,7 @@ module ElasticRecord
     attr_accessor :disabled
     attr_accessor :model
     attr_accessor :partial_updates
+    attr_accessor :load_from_source
 
     def initialize(models)
       models = Array.wrap(models)
@@ -63,6 +64,16 @@ module ElasticRecord
 
     def enable!
       @disabled = false
+    end
+
+    def load_from_source(&block)
+      if block_given?
+        @load_from_source = true
+        yield
+        @load_from_source = false
+      else
+        @load_from_source
+      end
     end
 
     def real_connection
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -62,11 +62,12 @@ module ElasticRecord
       klass.current_elastic_search = previous
     end
 
-    def search_hits
-      search_results['hits']['hits']
-    end
-
     private
+
+      def search_hits
+        search_results['hits']['hits']
+      end
+
       def reset
         @search_results = @records = nil
       end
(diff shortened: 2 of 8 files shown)


lib/elastic_record/relation.rb
```diff
@@ -75,13 +75,21 @@ def search_results
...
+          end
         end
       end
 
-      def load_hits(ids)
-        scope = select_values.any? ? klass.select(select_values) : klass
-        scope.find(ids)
+      def load_hits
+        if klass.elastic_index.load_from_source
+           search_hits&.map { |hit| klass.new(hit['_source'].merge('id' => hit['_id'])) }
```
matthuhiggins: When are `search_hits` nil?

lib/elastic_record/index.rb
```diff
@@ -65,6 +66,16 @@ def enable!
...
 
+    def load_from_source(&block)
+      if block_given?
+        @load_from_source = true
+        yield
+        @load_from_source = false
+      else
+        @load_from_source
+      end
+    end
```
matthuhiggins: I'd rather think about the block style API a bit more. Do you need this yet?

bullfight:
One limitation to this is that it still requires the definition of at least two additional class methods.  We should decide if this is something to document or something that shouldn't be required.

```ruby
def self.primary_key
  'id'
end

def self.base_class
  self
end
```

matthuhiggins:
ElasticRecord::Relation expects ActiveRecord compatibility. That's completely fine.

lib/elastic_record/relation.rb
bullfight: Hrm, I think this is unnecessary.

lib/elastic_record/index.rb
bullfight: I'm only using it for the tests, but it isn't necessary

test/dummy/.env.test
```diff
@@ -1 +1 @@
-DATABASE_URL=postgres://localhost/elastic_record_test
+DATABASE_URL=postgres://localhost/elastic_record_subtest
```
bullfight: I'm not sure if this is required but this was the name of the database generated by `bundle exec rake app:db:setup`.  I wasn't able to get the tests to run until I changed this env config.

matthuhiggins: What version of rails is in your `Gemfile.lock`?

test/elastic_record/relation_test.rb
```diff
@@ -42,6 +42,19 @@ def test_to_a
...
+  def test_to_a_from_source
+    Widget.elastic_index.load_from_source = true
+
+    create_widgets [Widget.new(id: 5, color: 'red'), Widget.new(id: 10, color: 'blue')]
+
+    array = Widget.elastic_relation.to_a
+
+    assert_equal 2, array.size
+    assert array.first.is_a?(Widget)
+
```
matthuhiggins: `ensure`

lib/elastic_record/relation.rb
```diff
@@ -75,13 +75,21 @@ def search_results
...
+          end
         end
       end
 
-      def load_hits(ids)
-        scope = select_values.any? ? klass.select(select_values) : klass
-        scope.find(ids)
+      def load_hits
+        if klass.elastic_index.load_from_source
+           search_hits.map { |hit| klass.new(hit['_source'].merge('id' => hit['_id'])) }
```
ebarendt: `merge` is going to create a new hash for each record. You can probably use `update` instead here.

malept: or `merge!`. Not entirely sure which one's the alias based on the Ruby docs.

test/dummy/.env.test
bullfight: Rails 5.1.4

lib/elastic_record/relation.rb
bullfight: I was trying to avoid modifying the `hits` objects in place.  I think it would be acceptable to do so if we make `search_hits` a private method.  It looks like it not called outside of `ElasticRecord::Relation`.

bullfight:
> RE: about the block style, make a separate method such as `Widget.elastic_relation.loading_from_source(&block)`

Should `load_from_source` be a configuration on `elastic_index` or on `elastic_relation`?  While it is used only in `elastic_relation` we don't often make calls to `elastic_relation` directly.

bullfight:
Ah, there is also the problem that it isn't memoized

https://github.com/data-axle/elastic_record/blob/master/lib/elastic_record/searching.rb#L3-L5

PR #54, opened 2017-09-13
Support Automatic ID Generation

bullfight:
ElasticSearch supports [automatically generating an ID for documents](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/docs-index_.html#_automatic_id_generation) where id is not specified. To automatically generate an id with ElasticSearch the document must be sent as a `POST` request to the index.

This adds a condition to `ElasticRecord::Index::Documents#index_document` which will `POST` a document to the index if no id is specified.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -93,6 +93,7 @@ module ElasticRecord
       end
 
       def update_document(id, document, doctype: model.doctype, parent: nil, index_name: alias_name)
+        raise "Cannot update a document with empty id" if id.blank?
         params = {doc: document, doc_as_upsert: true}
 
         if batch = current_bulk_batch
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -54,7 +54,7 @@ module ElasticRecord
       def index_record(record, index_name: alias_name)
         unless disabled
           index_document(
-            record.send(record.class.primary_key),
+            record_primary_key(record),
             record.as_search_document,
             doctype: record.doctype,
             index_name: index_name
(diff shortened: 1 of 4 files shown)


lib/elastic_record/index/documents.rb
```diff
@@ -84,7 +84,11 @@ def index_document(id, document, doctype: model.doctype, parent: nil, index_name
           path = "/#{index_name}/#{doctype.name}/#{id}"
           path << "?parent=#{parent}" if parent
 
-          connection.json_put path, document
+          if id
+            connection.json_put path, document
+          else
+            connection.json_post path, document
+          end
```
matthuhiggins: The model should not be required to implement the `id` method. Use `#try(:id)`.

bullfight: I think what you intended with this was that the model shouldn't need to define `primary_key`.  The change above is on `index_document` which takes `id` as it's first argument.

matthuhiggins: That code was not shown, so I commented here.

Use try(:id). The new `record_primary_key` is useless, since that's what rails already does with `id`.

bullfight: Ah, I get it.

lib/elastic_record/index/documents.rb
```diff
@@ -65,7 +65,7 @@ def index_record(record, index_name: alias_name)
       def update_record(record, index_name: alias_name)
         unless disabled
           update_document(
-            record.send(record.class.primary_key),
+            record.try(:id),
```
bullfight: Should we raise an error on `update_record` if no ID is provided?

bullfight: Ah, this should fail on the `update_document` call anway

matthuhiggins:
Don't use `try` in update_record.

bullfight:
> I don't see the reason for this change in update_record:

That change is in `index_document` the split diff has lines collapsed and makes it look like it is in `update_record`

matthuhiggins:
Ah, silly me

PR #56, opened 2017-10-06
Support source options with batch methods

matthuhiggins:
`Relation::Batch` was misbehaving with `Index#load_from_source`. This PR fixes two issues:

Problem 1:

When `load_from_source=false`, source was still being loaded. This makes things _slow_!

Problem 2:

When `load_from_source=true`, the code tried to call `AR.where`, rather than respecting the option to use source.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -13,8 +13,8 @@ module ElasticRecord
       end
 
       def each_slice(&block)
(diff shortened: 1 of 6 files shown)


bullfight:
Maybe this test case should be re-written for `load_hits` for the `Hits` module

https://github.com/data-axle/elastic_record/blob/3184dbeb267c6220899bc79393827690db644bd0/test/elastic_record/relation_test.rb#L45-L56

PR #58, opened 2018-01-09
Add BulkDocuments concern and make bulk operations threadsafe

matthuhiggins:
Extract bulk operations into new `BulkDocuments` module. Some of the methods override methods in `Document`, calling super if not inside a bulk operation.

Further, use `Thread.current` to make bulk operations threadsafe. ElasticRecord is currently not threadsafe because it `Connection#bulk_stack` is a variable on the global instance of connection, which is shared across all threads.

### DIFF
--- a/lib/elastic_record/index/bulk_documents.rb
+++ b/lib/elastic_record/index/bulk_documents.rb
@@ -16,8 +16,7 @@ module ElasticRecord
       end
 
       def update_document(id, document, doctype: model.doctype, parent: nil, index_name: alias_name)
-        raise "Cannot update a document with empty id" if id.blank?
-        params = {doc: document, doc_as_upsert: true}
+        params = validate_and_extract_update_params(id, document)
 
         if batch = current_bulk_batch
           instructions = { _index: index_name, _type: doctype.name, _id: id, _retry_on_conflict: 3 }
--- a/lib/elastic_record/index/bulk_documents.rb
+++ b/lib/elastic_record/index/bulk_documents.rb
@@ -44,7 +44,7 @@ module ElasticRecord
       end
 
       def bulk(options = {})
-        connection.bulk_stack.push []
+        ElasticRecord::Connection.bulk_stack.push []
 
         yield
 
@@ -54,7 +54,7 @@ module ElasticRecord
           verify_bulk_results(results)
         end
       ensure
-        connection.bulk_stack.pop
+        ElasticRecord::Connection.bulk_stack.pop
       end
 
       def bulk_add(batch, index_name: alias_name)
@@ -66,7 +66,7 @@ module ElasticRecord
       end
 
       def current_bulk_batch
-        connection.bulk_stack.last
+        ElasticRecord::Connection.bulk_stack.last
       end
 
       private
--- a/lib/elastic_record/index/bulk_documents.rb
+++ b/lib/elastic_record/index/bulk_documents.rb
@@ -31,8 +31,7 @@ module ElasticRecord
       end
 
       def delete_document(id, doctype: model.doctype, parent: nil, index_name: alias_name)
-        raise "Cannot delete document with empty id" if id.blank?
-        index_name ||= alias_name
+        validate_doc_deletion(id, index_name)
 
         if batch = current_bulk_batch
           instructions = { _index: index_name, _type: doctype.name, _id: id, _retry_on_conflict: 3 }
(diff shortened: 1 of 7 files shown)


lib/elastic_record/index/bulk_documents.rb
```diff
@@ -0,0 +1,83 @@
+require 'active_support/core_ext/object/to_query'
+
+module ElasticRecord
+  class Index
+    module BulkDocuments
+      def index_document(id, document, doctype: model.doctype, parent: nil, index_name: alias_name)
+        if batch = current_bulk_batch
+          instructions = { _index: index_name, _type: doctype.name, _id: id }
```
ebarendt: Add similar ID verification for nil?

ebarendt: And retry?

lib/elastic_record/index/bulk_documents.rb
```diff
@@ -0,0 +1,83 @@
...
+          instructions[:parent] = parent if parent
+
+          batch << { index: instructions }
+          batch << document
+        else
+          super
```
bullfight: This calls `ElasticRecord::Index::Documents#index_record`?  I see that this supports the current `index.bulk {}` api, but I'm not a huge fan of the overloading/indirection. I would prefer to see these methods split out into a `BulkDocument` instance which is returned when calling `bulk` e.g.

```ruby
index.bulk do |bulk_document|
  bulk_document.index_document '5', color: 'green'
end
```

I find it easier to comprehend and debug a call to a class instance than to a method defined in multiple places.

This would also allow for a threadsafe by default `bulk_stack` that lives on the `BulkDocument` instance instead of `ElasticRecord::Connection`.

lib/elastic_record/index/bulk_documents.rb
matthuhiggins: `id` is not required. Id is optional with the Index Document API.

`retry_on_conflict` is only available with the Update API.

matthuhiggins:
> Making bulk_stack threadsafe is a big improvement but maybe the API is a little too cute. Is there a case for it that I am missing?

The public API is unchanged. Are you referring to the public API, or the internal one?

bullfight:
> The public API is unchanged. Are you referring to the public API, or the internal one?

The implementation of the internal API is in service of maintaining the public API.  I think that the internal API could be simpler, but it would require changing the public API.

lib/elastic_record/index/bulk_documents.rb
matthuhiggins: > This would also allow for a threadsafe by default `bulk_stack` that lives on the `BulkDocument` instance instead of `ElasticRecord::Connection`.

This cannot work, because each model has their own index. The bulk saves must be tracked across all models. For example:
```
bulk do
  save_widget
  save_customer
end
```

The requirement to track the `bulk_document` variable does make sense, but we have code 10 to 20 method calls deep that rely on being inside of a transaction and a bulk index. Introducing the requirement to pass a `bulk_document` variable along through every method call, and requiring a method to even know it is in a bulk write, would introduce a lot of code into the application.

As for the design pattern and indirection, I agree that secondary change of refactoring out the bulk methods is is not a lot better. I was trying to group together code and tests that relate to bulk actions separately from the standard methods. ActiveRecord does this a lot, though the case is often different because it's not in the format of `if bulk new_code else super end`.

There are many examples of this in AR:
https://github.com/rails/rails/blob/master/activerecord/lib/active_record/validations.rb#L45-L53

matthuhiggins:
> > The public API is unchanged. Are you referring to the public API, or the internal one?

> The implementation of the internal API is in service of maintaining the public API.  I think that the internal API could be simpler, but it would require changing the public API.

Try typing out your public API suggestion in a typical use case in our application.

lib/elastic_record/index/bulk_documents.rb
bullfight: Ah! Thank you for this explanation. Thinking about it's use as similar to an `ActiveRecord::Base#transaction` makes the requirement for this design much more obvious.

dcromer:
I like how the `Documents` module has become simpler, but this is offset by the confusion of BulkDocuments overriding every method in `Documents` and going

```
if bulk = current_bulk_stack
  #do da bulk
else
  super
end
```

If you want to keep the interface,

```
index.bulk do
  index.index_record ...
end
```

I think it would be simpler to just do

```
# ::Documents
def index_record
  if bulk_stack = current_bulk_stack
    _index_bulk(...)
  else
    _index_non_bulk(...)
  end
end
```

And then I don't need to think about what `super` is.

matthuhiggins:
Canceling this and just fixing thread safety with https://github.com/data-axle/elastic_record/pull/59. Thanks for the feedback.

PR #62, opened 2018-02-15
Support includes and use active record for testing

matthuhiggins:
This PR accomplishes two things:
* Adds support for `.includes`. This allows for `Widget.includes(:factory).filter(color: 'red')`, and the factory relationship will be preloaded.
* Replace "mock" Widget with a real `Widget` model that has a table. This allows for deeper testing and increases authenticity of tests.

TODO:
- [x] Fix `test_as_partial_update_document`.
- [x] Add a test to assert that `includes` successfully preloads records.


ebarendt:
Nice. So much code to clean up.

bullfight:
This is great. I am impressed that so little code was required to support this feature.

PR #63, opened 2018-02-16
A more generalized pattern for ar relation integration

matthuhiggins:


lib/elastic_record/relation/value_methods.rb
```diff
@@ -1,6 +1,6 @@
 module ElasticRecord
   class Relation
-    MULTI_VALUE_METHODS  = [:extending, :filter, :includes, :order, :select, :aggregation]
+    MULTI_VALUE_METHODS  = [:extending, :filter, :order, :aggregation]
```
ebarendt: I haven't looked - do you need to add `:joins` here since you included it above?

PR #65, opened 2018-03-12
ES6 Upgrade 

kunruh9:
Current state of the ES6 upgrade changes. Primarily addresses [removal of mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html).

- [x] Additional minor changes still need to be looked over and determined if any additional code tweaks are needed (there is a [long list](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-6.0.html#_also_see)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ sudo: required
 dist: trusty
 before_install:
 - echo "deb https://artifacts.elastic.co/packages/6.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-6.x.list
-- sudo apt-get install elasticsearch
-- sudo service elasticsearch start; sleep 10
+- sudo apt-get update && sudo apt-get install -y elasticsearch
+- sudo service elasticsearch start
 before_script:
 - cp test/dummy/.env.example test/dummy/.env
-- curl localhost:9200
+- until ping -c1 http://localhost:9200 &>/dev/null; do :; done
 - bundle exec rake app:db:setup
 - bundle exec rake app:index:reset
 services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 rvm: 2.4.1
 cache: bundler
-sudo: false
+sudo: required
 dist: trusty
 before_install:
 - echo "deb https://artifacts.elastic.co/packages/6.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-6.x.list
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ dist: trusty
 before_install:
 - echo "deb https://artifacts.elastic.co/packages/6.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-6.x.list
 - sudo apt-get install elasticsearch
+- sudo service elasticsearch start
 before_script:
 - cp test/dummy/.env.example test/dummy/.env
 - curl localhost:9200
 - bundle exec rake app:db:setup
 - bundle exec rake app:index:reset
 services:
-- elasticsearch
 - postgresql
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ dist: trusty
 before_install:
 - echo "deb https://artifacts.elastic.co/packages/6.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-6.x.list
 - sudo apt-get install elasticsearch
-- sudo service elasticsearch start
+- sudo service elasticsearch start; sleep 10
 before_script:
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -1,18 +1,15 @@
...
+- sudo apt-get install elasticsearch
+- sudo service elasticsearch start; sleep 10
 before_script:
 - cp test/dummy/.env.example test/dummy/.env
 - curl localhost:9200
 - bundle exec rake app:db:setup
 - bundle exec rake app:index:reset
 services:
-- elasticsearch
 - postgresql
```
matthuhiggins: You could add something like this in the `before_script`, which replaces uses of both `sleep` and `curl`:

https://serverfault.com/questions/42021/how-to-ping-in-linux-until-host-is-known

ping defaults to 10s but that can be changed to 1s.

lib/elastic_record/doctype.rb
```diff
@@ -10,18 +10,6 @@ class Doctype
...
-      properties: {
-        query: {
-          type: "percolator"
-        }
-      }
-    }
-
-    def self.percolator_doctype
-      new('queries', PERCOLATOR_MAPPING)
-    end
```
kunruh9: Multi-doctype no longer allowed. Percolators should now share the same doctype with an additional `query: {type: percolator}` field added to the mapping, or (more ideally), be contained in their own separate index.

test/dummy/app/models/widget.rb
```diff
@@ -7,16 +7,13 @@ class Widget < ActiveRecord::Base
 
   self.doctype.mapping[:properties].update(
     'name' => {
-      type: 'string', index: 'not_analyzed',
-      fields: {
-        analyzed: {type: 'string', index: 'analyzed'}
-      }
+      type: 'text',
```
kunruh9: This seemed to be the closest equivalent mapping I could determine for this property. I couldn't find any documentation (for previous or current versions) on the `field: analyzed: {...}` mapping.

test/dummy/app/models/widget_query.rb
```diff
@@ -5,6 +5,14 @@ class WidgetQuery
...
   self.percolates_model = Widget
 
+  doctype.mapping = {
+    properties: {
+      name: { type: 'keyword' },
+      color: { type: 'keyword' },
+      query: { type: 'percolator' }
+    }
+  }
+
```
kunruh9: Without defining mapping, the Percolator Model in the test suite was being created with a very confusing, and seemingly incorrect, mapping. See image below:

screen shot 2018-03-13 at 1 23 13 pm

I think explicit mapping may be required a majority of the time. Determining mapping dynamically could probably use some improvement.

kunruh9:
@matthuhiggins I believe this should be done. I looked through the other ES6 changes and they all looked like changes that only require potential changes to code that uses this library.

Let me know if you see anything that looks questionable.

test/dummy/app/models/widget.rb
matthuhiggins: text is good.

ebarendt: Except it looks like we lose tests for `name` (keyword) vs `name.analyzed` (text), right?

matthuhiggins:
This looks good. See if it works in the application.

test/dummy/app/models/widget.rb
kunruh9: Possibly. I updated the mapping, probably more accurate now.

kunruh9:
@matthuhiggins Finally got it working now with content_system and axle_place. Encountered some really hard-to-track-down bugs related to key duplications in mapping.

Should all be good now.

lib/elastic_record/doctype.rb
```diff
@@ -18,12 +18,12 @@ class Doctype
...
     def initialize(name, mapping = DEFAULT_MAPPING.deep_dup)
       @name = name
-      @mapping = mapping
+      @mapping = mapping.deep_symbolize_keys
       @analysis = {}
     end
 
     def mapping=(custom_mapping)
-      mapping.deep_merge!(custom_mapping)
+      mapping.deep_merge!(custom_mapping.deep_symbolize_keys)
```
matthuhiggins: Is symbolizing keys necessary?

kunruh9: Well, _strictly_ speaking, no. But it's certainly a lot safer...I found that our Places mapping had a mixture of string keys and symbol keys before creating the index. And when additional mapping is added, `deep_merge` treats `"key" => {...}` and `:key => {...}` as different and wont merge them. But when converting to JSON and submitting it as a request body to ES, those keys will be the same and be duplicated.

In < ES 6, this was fine, it would just take the most recent duplicated key and use that value. But in ES 6+, duplicate keys will return an error.

PR #66, opened 2018-03-23
[WIP] ES6 Compatibility Upgrade

kunruh9:
The primary change here is how `mapping_body` behaves. Both ES5 and ES6 versions will now us a single doctype mapping strategy, regardless if multiple models were used to initialize the index. Since both versions support this, the approach alleviates the need have the `doctype.name` be dynamic depending on the ES version.

As an alternative, the mapping also adds a `custom_doctype` field to the mapping which would allow implementations to filter results by a doctype if multiple models share the same index. However, models are not automatically index with the `custom_doctype` set appropriately...I was struggling to get this working properly.

Ideally, I think we'd want something like this to be possible:

```ruby
ElasticRecord::Index.new([Widget, Product]).create_and_deploy

Widget.create color: 'red'
Product.create color: 'red'

Widget.elastic_search.filter(color: 'red').count # only returns one result -- document with custom_doctype == 'widget' and 'color' == red

Widget.elastic_index == Product.elastic_index # also should have the same index reference
```

When attempting to get the above working, I ran into some issues and am still investigating.

### DIFF
--- a/lib/elastic_record/index/manage.rb
+++ b/lib/elastic_record/index/manage.rb
@@ -76,7 +76,7 @@ module ElasticRecord
       end
 
       def all_names
-        connection.json_get("/#{alias_name}/_mapping/#{model.doctype.name}/").keys
+        connection.json_get("/#{alias_name}/_mapping").keys
       rescue
         # TODO: In ES 1.4, this returns empty rather than a 404
         []
--- a/lib/elastic_record/index/mapping.rb
+++ b/lib/elastic_record/index/mapping.rb
@@ -14,9 +14,11 @@ module ElasticRecord
       end
 
       def mapping_body
-        doctypes.each_with_object({}) do |doctype, result|
-          result[doctype.name] = doctype.mapping
+        mapping = doctypes.each_with_object({}) do |doctype, result|
+          result.deep_merge! doctype.mapping
         end
+
+        { model.doctype.name.to_sym => mapping }
       end
     end
   end
--- a/lib/elastic_record/percolator_model.rb
+++ b/lib/elastic_record/percolator_model.rb
@@ -27,7 +27,7 @@ module ElasticRecord
         query = {
           "percolate" => {
             "field"         => "query",
-            "document_type" => percolates_model.doctype.name,
+            "document_type" => doctype.name,
             "document"      => document
           }
         }
--- a/lib/elastic_record/percolator_model.rb
+++ b/lib/elastic_record/percolator_model.rb
@@ -20,14 +20,14 @@ module ElasticRecord
       end
 
       def doctype
-        @doctype ||= Doctype.percolator_doctype
+        @doctype ||= Doctype.new(percolates_model.doctype.name, Doctype::PERCOLATOR_MAPPING)
       end
 
       def percolate(document)
         query = {
           "percolate" => {
             "field"         => "query",
-            "document_type" => doctype.name,
+            "document_type" => percolates_model.doctype.name,
             "document"      => document
           }
         }
--- a/lib/elastic_record/percolator_model.rb
+++ b/lib/elastic_record/percolator_model.rb
@@ -20,7 +20,11 @@ module ElasticRecord
       end
 
       def doctype
-        @doctype ||= Doctype.new(percolates_model.doctype.name, Doctype::PERCOLATOR_MAPPING)
+        @doctype ||=
+          begin
+            doctype_name = check_mapped_doctype_name || percolates_model.doctype.name
+            Doctype.new(doctype_name, Doctype::PERCOLATOR_MAPPING)
+          end
       end
 
       def percolate(document)
@@ -34,6 +38,26 @@ module ElasticRecord
 
         elastic_search.filter(query).limit(5000)
       end
+
+      private
+
+        def check_mapped_doctype_name
+          mapped_doctypes = elastic_connection.json_get("/#{base_class.name.demodulize.underscore.pluralize}/_mapping").values.first['mappings']
+
+          if mapped_doctypes.keys.count > 1
+            mapped_percolator_doctype(mapped_doctypes)
+          end
+        rescue ElasticRecord::ConnectionError
+          nil
+        end
+
+        def mapped_percolator_doctype(doctypes_map)
+          doctypes_map.each_pair do |doctype_name, mapping|
+            mapping['properties'].each_value do |value|
+              return doctype_name if value == {"type" => "percolator"}
+            end
+          end
+        end
     end
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,22 @@ rvm: 2.4.1
 cache: bundler
 sudo: false
 dist: trusty
-addons:
-  apt:
-    sources:
-    - elasticsearch-5.x
-    packages:
-    - elasticsearch
+
+before_install:
+- wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.tar.gz
+- tar -xzf elasticsearch-${ES_VERSION}.tar.gz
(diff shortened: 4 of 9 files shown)


lib/elastic_record/connection.rb
```diff
@@ -17,6 +19,10 @@ def initialize(servers, options = {})
       self.bulk_stack         = []
     end
 
+    def elastic_version
+      @elastic_version ||= get_elastic_version
+    end
```
kunruh9: Actually might not need this anymore. I think I took all the conditional checks out.

lib/elastic_record/as_document.rb
```diff
@@ -1,7 +1,7 @@
 module ElasticRecord
   module AsDocument
     def as_search_document
-      doctype.mapping[:properties].each_with_object({}) do |(field, mapping), result|
+      doctype.mapping[:properties].each_with_object({}.with_indifferent_access) do |(field, mapping), result|
```
malept: There's gotta be a better way to deal with mixed hash key types instead of peppering `.with_indifferent_access` everywhere.

lib/elastic_record/index.rb
```diff
@@ -40,9 +40,9 @@ class Index
     attr_accessor :load_from_source
 
     def initialize(models)
-      models = Array.wrap(models)
+      models = Array.wrap models
```
malept: Is removing the parentheses really necessary?

kunruh9: Nah, that was a side effect of multiple iterations of this file and ultimately ending up back at the same functionality as before. I'll add the parentheses back in to reduce changes.

lib/elastic_record/as_document.rb
kunruh9: Originally, I was using `deep_symbolize_keys` in the accessor methods for `mapping`. But using `with_indifferent_access` was per @matthuhiggins. I'm open to better ways though.

lib/elastic_record/index/mapping.rb
```diff
@@ -14,9 +14,19 @@ def get_mapping(index_name = alias_name)
...
+            properties: {
+              custom_doctype: { type: 'keyword' } }
+          }
+        }
+
+        doctypes.each do |doctype|
+          map[ElasticRecord::Config.default_doctype_name.to_sym].deep_merge! doctype.mapping
         end
+
+        map
```
malept: Perhaps:

```ruby
def mapping_body
  doctype_mappings = {
    properties: {
      custom_doctype: { type: 'keyword' }
    }
  }
  doctypes.each do |doctype|
    doctype_mappings.deep_merge!(doctype.mapping)
  end

  {
    ElasticRecord::Config.default_doctype_name.to_sym => doctype_mappings
  }
end
```

lib/elastic_record/as_document.rb
matthuhiggins: I meant for `with_indifferent_access` on the `mapping` and `setting` instances, not on calls to the instance.

lib/elastic_record/index.rb
```diff
@@ -80,7 +80,7 @@ def real_connection
 
     def get(end_path, doctype = nil, json = nil)
       path = "/#{alias_name}"
-      path += "/#{doctype.name}" if doctype
+      path += "/#{ElasticRecord::Config.default_doctype_name}" if doctype
```
ebarendt: This doesn't seem to match with the `if` at the end or the argument passed in. Is the argument left here (and other places) for api compatibility? If so, when does it get removed?

kunruh9: Yes, it is for api compatibility.
As for when it gets removed, that's a good question. These changes force all models to be mapped in an index with a single doctype (`doc` by default). Because of this, it's no longer valid to use a dynamic doctype the ES rest API. The complication is that `get` isn't necessarily always used for queries. So there isn't a clean way to inject the given doctype name into the `json` body to force the request to only return documents of the given doctype using `custom_doctype` field (see [mapping.rb](https://github.com/data-axle/elastic_record/blob/es6/lib/elastic_record/index/mapping.rb#L19)).

Right now, the expectation is that implementations of the library should manually set the `custom_doctype` value in query/filter requests. Though that isn't very intuitive, so an automatic solution that behaves similar to the current release of this library would be better. I found that to be difficult though since there are a lot of places, especially [documents.rb](https://github.com/data-axle/elastic_record/blob/es6/lib/elastic_record/index/documents.rb), that would need some conditional injections to the body.

ebarendt: How does this work with code that is passing in a doctype currently? The old index and the new won't be in the same place.

kunruh9: Hmm, yeah, it doesn't work for that case :( Sigh, complications of trying to upgrade a library with backwards compatibility between major versions.

I'm going to have to brainstorm this...this multi-version support is throwing me for a loop.

lib/elastic_record/config.rb
```diff
@@ -34,6 +37,8 @@ def settings=(settings)
         if settings['scroll_keep_alive'].present?
           self.scroll_keep_alive = settings['scroll_keep_alive']
         end
+
+        self.default_doctype_name = settings['default_doctype_name'] if settings['default_doctype_name'].present?
```
matthuhiggins: No reason for this. The app can write this in ruby without loading yml, etc.

lib/elastic_record/index.rb
matthuhiggins: I think it's better to approach it from a "does the index have multiple doc types" versus "what version of ES is it".

kunruh9:
@matthuhiggins @malept @ebarendt 

I spent a long time trying to figure out how to expand this library to support indexes created with multiple models (a.k.a multiple doctypes) while still having search/query methods automatically 'filter' by each doctype it contains. Ultimately what I found is that this isn't a feature implemented by the current library (with exception to percolators). See af8ff460f8dbb77c5ab7c0f45f57c76842344b21 for my demonstration of this.

Once I realized this, the actual changes needed were simplified greatly. The only change needed is how the mapping body is built when creating the index. All the internal query logic already uses the first model doctype for a given index as the doctype, so nothing needs to change there. To keep things simple, I nixed the global generic doctype that I initially had. Index mappings will continue to use the first model doctype name as the index doctype upon creation, which is compatible with ES v5 and v6.

lib/elastic_record/index/manage.rb
```diff
@@ -76,7 +76,7 @@ def aliased_names
       end
 
       def all_names
-        connection.json_get("/#{alias_name}/_mapping/#{model.doctype.name}/").keys
+        connection.json_get("/#{alias_name}/_mapping").keys
```
kunruh9: This change doesn't really have anything to do with the ES6 upgrade. I found that the endpoint actually wouldn't return **all** indexes that the given alias represents, only the first one. Removing the doctype seemed to fix this.

lib/elastic_record/percolator_model.rb
```diff
@@ -34,6 +38,26 @@ def percolate(document)
...
       end
+
+      private
+
+        def check_mapped_doctype_name
+          mapped_doctypes = elastic_connection.json_get("/#{base_class.name.demodulize.underscore.pluralize}/_mapping").values.first['mappings']
+
+          if mapped_doctypes.keys.count > 1
+            mapped_percolator_doctype(mapped_doctypes)
+          end
```
matthuhiggins: Why not use `multiple_doctypes?`

kunruh9: Because I'd have to reference `elastic_index` which I found creates an infinite loop. Index initializes which calls `doctype` on all models passed which again calls `check_mapped_doctype_name` which would call `elastic_index` again.

I realize this is all kind of messy, still a WIP. I just wanted to see if this solution would actually work before cleaning it up.

PR #69, opened 2018-04-20
Handle the happy path of nested bulk blocks

malept:
Turns out I didn't actually finish #68.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -164,18 +164,23 @@ module ElasticRecord
       end
 
       def bulk(options = {})
-        return if current_bulk_batch
-        connection.bulk_actions = []
+        if current_bulk_batch
+          yield
+        else
+          begin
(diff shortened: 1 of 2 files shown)


lib/elastic_record/index/documents.rb
```diff
@@ -164,18 +164,23 @@ def scroll(scroll_id, scroll_keep_alive)
...
+            yield
+
+            if current_bulk_batch.any?
+              body = current_bulk_batch.map { |action| "#{JSON.generate(action)}\n" }.join
+              results = connection.json_post("/_bulk?#{options.to_query}", body)
+              verify_bulk_results(results)
+            end
+          ensure
+            connection.bulk_actions = nil
+          end
```
malept: I'd like to move the `begin...ensure...end` into its own private method but I don't know what to name it right now.

ebarendt: `run_new_bulk_batch`?

PR #71, opened 2018-06-04
Range type

matthuhiggins:
Adds support for elasticsearch's [range datatype](https://www.elastic.co/guide/en/elasticsearch/reference/current/range.html).

### DIFF
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -47,9 +47,21 @@ module ElasticRecord
     end
 
     def value_for_elastic_search_range(range)
-      gte = range.begin unless range.begin == -Float::INFINITY
-      lte = range.end unless range.end == Float::INFINITY
+      ordered = ordered_range(range)
+
+      gte = ordered.begin unless ordered.begin == -Float::INFINITY
+      lte = ordered.end unless ordered.end == Float::INFINITY
       {'gte' => gte, 'lte' => lte}
     end
+
+    private
+
+      def ordered_range(range)
+        if range.begin > range.end
+          range.end..range.begin
+        else
+          range
+        end
+      end
   end
 end
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -47,21 +47,12 @@ module ElasticRecord
     end
 
     def value_for_elastic_search_range(range)
-      ordered = ordered_range(range)
+      if range.begin <= range.end
+        gte = range.begin unless range.begin == -Float::INFINITY
+        lte = range.end unless range.end == Float::INFINITY
 
-      gte = ordered.begin unless ordered.begin == -Float::INFINITY
-      lte = ordered.end unless ordered.end == Float::INFINITY
-      {'gte' => gte, 'lte' => lte}
-    end
-
-    private
-
-      def ordered_range(range)
-        if range.begin > range.end
-          range.end..range.begin
-        else
-          range
-        end
+        {'gte' => gte, 'lte' => lte}
       end
+    end
   end
 end
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -47,12 +47,10 @@ module ElasticRecord
     end
 
     def value_for_elastic_search_range(range)
-      if range.begin <= range.end
-        gte = range.begin unless range.begin == -Float::INFINITY
-        lte = range.end unless range.end == Float::INFINITY
+      gte = range.begin unless range.begin == -Float::INFINITY
+      lte = range.end unless range.end == Float::INFINITY
 
-        {'gte' => gte, 'lte' => lte}
-      end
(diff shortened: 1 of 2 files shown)


lib/elastic_record/as_document.rb
```diff
@@ -16,27 +16,40 @@ def as_partial_update_document
...
     end
+
+    def value_for_elastic_search_object(object)
+      object.respond_to?(:as_search_document) ? object.as_search_document : object
+    end
+
+    def value_for_elastic_search_range(range)
+      gte = range.begin unless range.begin == -Float::INFINITY
+      lte = range.end unless range.end == Float::INFINITY
+      {'gte' => gte, 'lte' => lte}
```
dcromer: If the beginning or end are infinity, then our document value is `nil`.  Is this acceptable for an ES range?  The documentation does not mention anything about nil values.

dcromer: Ruby ranges can be specified in reverse:  

```
record = SpecialFieldsModel.new(meta: ['whatever'], book_length: 250..0)
```

This results in `{"gte" => 250, "lte" => 0}`.  Is that expected?

dcromer: OK, some more clarity on this:

1) If gte/lte is nil, _or one or both keys are omitted entirely_, the document indexes successfully, and the range is unbounded on that key(s).

2) If gte > lte, the document fails to index.

```
PUT /range_index/_doc/3 
{
  "expected_attendees" : { 
		"gte": 20,
		"lte": 10
  }
}
=> 
{
	"error": {
		"root_cause": [
			{
				"type": "mapper_parsing_exception",
				"reason": "failed to parse [expected_attendees]"
			}
		],
		"type": "mapper_parsing_exception",
		"reason": "failed to parse [expected_attendees]",
		"caused_by": {
			"type": "illegal_argument_exception",
			"reason": "min value (20) is greater than max value (10)"
		}
	},
	"status": 400
}
```

lib/elastic_record/as_document.rb
```diff
@@ -16,27 +16,52 @@ def as_partial_update_document
...
+
+    private
+
+      def ordered_range(range)
+        if range.begin > range.end
+          range.end..range.begin
+        else
+          range
+        end
+      end
```
matthuhiggins: This is silly. Why support this edge case? Is there a use case for it?

matthuhiggins: And if required, you could use min/max with fewer lines of code in the method above :)

dcromer: All elasticsearch ranges are ruby ranges.  Some ruby ranges are elasticsearch ranges.

Should it be possible to initialize a document with values that won't serialize?

> you could use min/max with fewer lines of code in the method above :)

Surprisingly you can't when `begin > end`!  [Range#min](https://ruby-doc.org/core-2.2.0/Range.html#method-i-min)

matthuhiggins: What test case shows this problem?

dcromer: > What test case shows this problem?

https://github.com/data-axle/elastic_record/pull/71/files#diff-32fef10a3489f3c4d2fd0ccf18043c8cR66

test/elastic_record/as_document_test.rb
```diff
@@ -48,17 +50,33 @@ def author
...
-    def meta
-      { some: "value" }
-    end
   end
 
   def test_as_search_document_with_special_fields
-    doc = SpecialFieldsModel.new.as_search_document
+    record = SpecialFieldsModel.new(meta: { some: "value" }, book_length: 250..500)
+
+    doc = record.as_search_document
```
matthuhiggins: Duplicate tests below. delete this one.

lib/elastic_record/as_document.rb
matthuhiggins: I see. Change superstore to make the range nil if begin > end. Allow ES to crash here if the data is bad.

lib/elastic_record/as_document.rb
```diff
@@ -16,27 +16,43 @@ def as_partial_update_document
...
         value
       end
     end
+
+    def value_for_elastic_search_object(object)
+      object.respond_to?(:as_search_document) ? object.as_search_document : object
+    end
+
+    def value_for_elastic_search_range(range)
+      if range.begin <= range.end
```
matthuhiggins: Delete this. Allow it to crash.

matthuhiggins: The crash will never happen though because ORM layer will never create the scenario.

dcromer: elastic_record does not depend on any ORM layer besides active_model.  You are correct that superstore will prevent this, though.

dcromer: removed in 8fc5191ec2216707740d5bd5148bc0dd9e158080

tests reflect expected behavior

PR #73, opened 2018-07-06
Nested mappings define as_search_document

ebarendt:
Before:
Nested children used a method to define `as_search_document`. In our implementation, that returned everything that was indexed. Fine for most things, except the new range type.

Now:
Nested child objects don't need to implement `as_search_document`. This will automatically traverse the mappings and do the same thing as the parent document.

Any methods that override `as_search_document` need to be updated to take an argument, even if they choose to ignore it.

### DIFF
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -1,8 +1,8 @@
 module ElasticRecord
   module AsDocument
-    def as_search_document(mappings = doctype.mapping[:properties])
-      mappings.each_with_object({}) do |(field, mapping), result|
-        value = value_for_elastic_search field, mapping, mappings
+    def as_search_document(mapping_properties = doctype.mapping[:properties])
+      mapping_properties.each_with_object({}) do |(field, mapping), result|
+        value = value_for_elastic_search field, mapping, mapping_properties
 
         unless value.nil?
           result[field] = value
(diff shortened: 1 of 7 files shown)


lib/elastic_record/as_document.rb
```diff
@@ -1,8 +1,9 @@
 module ElasticRecord
   module AsDocument
-    def as_search_document
-      doctype.mapping[:properties].each_with_object({}) do |(field, mapping), result|
-        value = value_for_elastic_search field, mapping
+    def as_search_document(mappings = doctype.mapping[:properties])
+      (mappings || []).each_with_object({}) do |(field, mapping), result|
+        nested = mappings.dig(field, :properties)
```
dcromer: Is it necessary to get the nested mapping out ahead of time (even when it does not exist?) . Only `value_for_elasticsearch` cares about this, and only once it's definitely a nested mapping.

dcromer:
> Any methods that override as_search_document need to be updated to take an argument, even if they choose to ignore it.

I don't think that's true?

ebarendt:
> I don't think that's true?

They do or ruby throws an error because you're passing an argument to a method that doesn't take it.

lib/elastic_record/as_document.rb
ebarendt: but the full `mappings` don't get passed into that method. I'll see what else can be done to clean things up though.

dcromer:
duh

lib/elastic_record/as_document.rb
ebarendt: cleaned up a little.. only thing to do is pass the mappings down, but then only the thing that needs it has to dig in.

ebarendt:
> Doesn't each object (and nested object) know its own mapping?

No.

ebarendt:
@matthuhiggins do you have a few minutes to talk about this?

PR #74, opened 2018-07-09
Allow settings to update to be passed in

ebarendt:
*Problem*
`update_settings` sends the output of the `settings` method which contains things that cannot be updated after index creation.

*Solution*
Allow the settings to be passed in. Also sets them using `settings=` so that the internal representation stays correct.

### DIFF
--- a/lib/elastic_record/index/settings.rb
+++ b/lib/elastic_record/index/settings.rb
@@ -18,8 +18,9 @@ module ElasticRecord
           end
       end
 
-      def update_settings(index_name = alias_name)
-        connection.json_put "/#{index_name}/_settings", settings
+      def update_settings(index_name = alias_name, settings: nil)
+        self.settings = settings if settings
+        connection.json_put "/#{index_name}/_settings", settings || self.settings
       end
 
       def analysis_body


lib/elastic_record/index/settings.rb
```diff
@@ -18,8 +18,9 @@ def settings
           end
       end
 
-      def update_settings(index_name = alias_name)
-        connection.json_put "/#{index_name}/_settings", settings
+      def update_settings(index_name = alias_name, settings: nil)
+        self.settings = settings if settings
```
matthuhiggins: Can this be defaulted in the method signature? `settings: default_settings`

ebarendt: No, the point is really to call `settings=` with the supplied values so the local instance variable is up-to-date. `settings=` has logic for merging into the entire set of settings for the index.

ebarendt:
Fixed up a little - I think this is all I need.

lib/elastic_record/index/settings.rb
```diff
@@ -18,7 +18,8 @@ def settings
           end
       end
 
-      def update_settings(index_name = alias_name)
+      def update_settings(index_name = alias_name, settings: self.settings)
+        self.settings = settings
```
matthuhiggins: erm... Calling the method changes the state of the object?

ebarendt: Yes - see the code above. It calls `deep_merge!` so that multiple calls to `elastic_index.settings=` work, I guess.

matthuhiggins: Right, but this is really bad. The index instance is a global object, and `update_settings` can be called through the lifetime of an app....

ebarendt: For the settings I have in mind, it's harmless. But I could have it not do this and then if you pass in settings, those are the ones that are updated. Right now the alternative is that the only way for an app to do a partial settings update is to call `json_put /{index}/_settings` directly.

ebarendt: Fixed and updated the PR description.

PR #75, opened 2018-07-12
Fix update_mapping crash. PUT each mapping type separately

dcromer:


dcromer:
ah, no, this is wrong

PR #78, opened 2018-08-14
Expose methods for getting the raw results of a query

ebarendt:
For some upcoming features, I would like to access the raw ES results.

It will help with:
1) Finding search scores
2) Getting the name of the query that matched

### DIFF
--- a/lib/elastic_record/relation/hits.rb
+++ b/lib/elastic_record/relation/hits.rb
@@ -17,22 +17,18 @@ module ElasticRecord
         hits.map { |hit| hit['_id'] }
       end
 
-      private
+      def search_hits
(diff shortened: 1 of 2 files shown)


ebarendt:
Alternately: introduce some objects to wrap the raw responses.

matthuhiggins:
Seems fine

I think it's premature to wrap with an object. Are you thinking of something like `ActiveRecord::Result`?

ebarendt:
Something like that maybe. But I'd also want the Results wrapped in something because there's info there (like max_score). I think raw is the way to go for now.

matthuhiggins:
A result object should be lightweight.

PR #79, opened 2018-10-03
Remove doctypes

dcromer:
# Problem

In ElasticRecord 5.x, we (I) added `ElasticRecord::Doctype` to abstract multiple mapping types.  In 6.x, they announced that multiple mapping types would be removed, rendering the Doctype abstraction useless.

https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html

# Solution

Remove Doctype.  Add `ElasticRecord::Index#type` to facilitate migrating mapping type names from `<lowercase_model_name>` to `_doc`.

# Upgrading Instructions

Set `ElasticRecord::Index#type` as needed to continue working with indexes that use  a non-`_doc` mapping name.

E.x.
```
class Widget
  elastic_index.type = 'widget' # preserves current mapping name behavior
end
```


matthuhiggins:
It looks like ElasticSearch now calls this a _mapping type_.

lib/elastic_record/model.rb
```diff
@@ -18,22 +18,14 @@ def inherited(child)
 
         if child < child.base_class
           child.elastic_index = elastic_index.dup
-          child.doctype = doctype.dup
+          child.elastic_index.model = child
```
dcromer: @matthuhiggins I think this is correct?  This seems to be consistent with how we use inherited models:  children get a separate index.

matthuhiggins: hmm depends how you want polymorphism to work. Would be more consistent with active record if they shared the index unless otherwise specified.

README.md
```diff
@@ -194,10 +196,23 @@ Mapping types will be removed in ElasticSearch 7.x.  To rename the default mappi
 
 ```ruby
 class Product
+  include ElasticRecord::Model
+
   elastic_index.mapping_type = 'product'
 end
 ```
 
+### Inheritence
```
matthuhiggins: Inheritance

test/elastic_record/model_test.rb
```diff
@@ -17,8 +17,11 @@ def test_elastic_index
     assert_equal Widget, index.model
   end
 
-  def test_elastic_index_inheritence
+  def test_elastic_index_inheritaence
```
malept: Unsure if typo or trolling 😄

dcromer:
@matthuhiggins I did change the concept to 'mapping type' per your suggestion, so I assume you approve of the rest of this PR.

I'll release a new major version after pulling, since this is a breaking change.

matthuhiggins:
Yes this is great.

PR #81, opened 2018-11-27
Fixes for find(ids)

ebarendt:
1) Don't raise an error on Model.find([]) - ActiveRecord doesn't.
2) Override the default ES limit of 10 records when finding by IDs.


ebarendt:
I'm not sure what to do about `Person.find` or `Person.first` yet. I could change how `load_from_source` works - but I'd want to delete the [unused block-form](https://github.com/data-axle/elastic_record/blob/41fb4d402a76ecc4d7458a78647ec5b18df93706/lib/elastic_record/index.rb#L68-L73).

PR #82, opened 2018-11-27
Make it easier to store a model entirely in ES

ebarendt:
Simply `include ElasticRecord::LoadFromSource` in the model.

No need for models to set the flag manually. Enables `Model.first`,
`find` and friends for easy lookup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,14 +224,14 @@ end
 
 ```
 
-Use `elastic_index.load_from_source = true` to configure an index without ActiveRecord.
+Call `load_from_source!` to configure an index without ActiveRecord. Finder methods will be
+delegated to the ElasticRecord module.
 
 ```ruby
 class Product
   include ActiveModel::Model
   include ElasticRecord::Record
-
-  self.elastic_index.load_from_source = true
+  elastic_index.load_from_source!
 end
 ```
 
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -42,7 +42,7 @@ module ElasticRecord
     def initialize(model)
       @model = model
       @disabled = false
-      @load_from_source = false
+      self.load_from_source = false
     end
 
     def initialize_copy(other)
@@ -66,15 +66,15 @@ module ElasticRecord
     end
 
     def load_from_source!
-      @load_from_source = true
+      self.load_from_source = true
       model.singleton_class.delegate :find, :find_by, :find_each, :find_in_batches, :first, to: :elastic_search
     end
 
     def loading_from_source(&block)
-      @load_from_source = true
+      self.load_from_source = true
       yield
     ensure
-      @load_from_source = false
+      self.load_from_source = false
     end
 
     def real_connection
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -65,6 +65,11 @@ module ElasticRecord
       @disabled = false
     end
 
+    def load_from_source!
+      @load_from_source = true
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -224,14 +224,13 @@ end
 
 ```
 
-Use `elastic_index.load_from_source = true` to configure an index without ActiveRecord.
+Include `ElasticIndex::LoadFromSource` to configure an index without ActiveRecord.
```
malept: Mention that finder methods on the model will get delegated to `ElasticRelation`.

lib/elastic_record/load_from_source.rb
```diff
@@ -0,0 +1,13 @@
...
+  module LoadFromSource
+    extend ActiveSupport::Concern
+
+    included do
+      elastic_index.load_from_source = true
+    end
+
+    module ClassMethods
+      delegate :find, :find_by, :find_each, :find_in_batches, :find_ids_in_batches, :first, to: :elastic_relation
+    end
```
matthuhiggins: I would like to see https://github.com/data-axle/content_system/blob/master/app/models/concerns/index_scope_methods.rb dealt with too.

README.md
```diff
@@ -224,14 +224,13 @@ end
...
 
 ```ruby
 class Product
   include ActiveModel::Model
   include ElasticRecord::Record
-
-  self.elastic_index.load_from_source = true
+  include ElasticRecord::LoadFromSource
```
matthuhiggins: 👎 Use a DSL.

ebarendt: Are opposed to the module or how it's included here?

I could add a `loaded_from_source` class method that calls `include`. Then we'd have:

```ruby
class Product
  include ActiveModel::Model
  include ElasticRecord::Record
  loaded_from_source
end
```

lib/elastic_record/load_from_source.rb
ebarendt: The whole module or the delegations?

README.md
bullfight: `elastic_index.load_from_source!`

I like this. It is very similar to the current `elastic_index.load_from_source = true` and the other index configurations we provide via `elastic_index`.

lib/elastic_record/load_from_source.rb
ebarendt: I gave it a try, but ran into problems with things that are ActiveRecord models but also a PercolatorModel. Pagination got screwed up.

ebarendt:
I think this is good now. I spent a bunch of time on trying to move the code out of that other module, but I need to move on.

README.md
matthuhiggins: Only opposed to adding the module as the way to get the behavior. I am suggesting that `elastic_index.load_from_source!` includes the module.

lib/elastic_record/index.rb
```diff
@@ -65,6 +65,11 @@ def enable!
       @disabled = false
     end
 
+    def load_from_source!
+      @load_from_source = true
```
matthuhiggins: Use `self.load_from_source`, in case the implementation does more than set a var?

ebarendt:
Thoughts on moving the `delegate :query, :filter, :aggregate` to here though?

matthuhiggins:
1. Some kind of way to specify which methods are delegated to, or groupings of methods:

* :finders
* :querying
* :calculations
* :batches

```ruby
elastic_index.active_record_overrides :finders, :batches
```

2. Shorten `Place.elastic_search` to `Place.es`. Require uses of `.es` to access methods such as `count`. Of course, `Place.filter(..).count` would continue to work.

PR #87, opened 2018-12-13
Add SearchHits wrapper

malept:
`ElasticRecord::SearchHits` is a class that wraps a search response.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -1,15 +1,11 @@
 require 'active_support/core_ext/object/to_query'
-require 'elastic_record/search_hits'
 
 module ElasticRecord
   class Index
     class ScrollEnumerator
-      include ElasticRecord::SearchHits
-
-      attr_reader :klass, :scroll_id, :keep_alive, :batch_size
+      attr_reader :scroll_id, :keep_alive, :batch_size
       def initialize(elastic_index, search: nil, scroll_id: nil, keep_alive:, batch_size:)
         @elastic_index = elastic_index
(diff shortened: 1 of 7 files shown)


matthuhiggins:
I like the idea of search hits wrapping the response, especially if that can be done without changing the API

lib/elastic_record/index/documents.rb
```diff
@@ -1,15 +1,19 @@
 require 'active_support/core_ext/object/to_query'
+require 'elastic_record/search_hits'
```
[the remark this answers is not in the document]
malept: Oops, I forgot to delete that.

lib/elastic_record/search_hits.rb
```diff
@@ -1,7 +1,7 @@
 module ElasticRecord
   class SearchHits
     attr_accessor :hits, :model
-    delegate(*Enumerable.instance_methods, to: :hits)
+    delegate(:[], *Enumerable.instance_methods, to: :hits)
```
matthuhiggins: I don't like this. Why are any delegations required?

malept: Oh, I can get rid of the `:[]`. A consumer's test was digging into the implementation details of a search hit. Delegating the Enumerable methods make it easier to convert existing code, and I think it makes sense, since it's a wrapper around a collection.

dcromer: What about

```
include Enumerable
delegate :each, to: :hits
```

malept: I just dropped it after talking to Matt IRL.

PR #88, opened 2018-12-20
[WIP] Skipping saving ES-only records to the database

ebarendt:
No errors happen, but doesn't run callbacks to index the record either.

### DIFF
--- a/lib/elastic_record/persistence_override.rb
+++ b/lib/elastic_record/persistence_override.rb
@@ -1,0 +1,7 @@
+module ElasticRecord
+  module PersistenceOverride
+    def save(*)
+      true
+    end
+  end
+end
--- a/lib/elastic_record.rb
+++ b/lib/elastic_record.rb
@@ -13,6 +13,7 @@ module ElasticRecord
   autoload :Lucene
   autoload :Model
   autoload :PercolatorModel
+  autoload :PersistenceOverride
   autoload :Relation
   autoload :Searching
   autoload :SearchHits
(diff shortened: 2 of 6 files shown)


matthuhiggins:
I think it will be more maintainable to add a `save_to_elastic_search` method.

Also, you can fix our app with proper use of FactoryBot, replacing the abusive use of `after_build`: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-methods-to-persist-objects

matthuhiggins:
@ebarendt What are you thoughts on `save_to_elastic_search`?

ebarendt:
I don't think it would help much. My goal with `save` was to make it transparent and not have to worry about whether the model was ES-only or not. With `save_to_elasticsearch`, the caller would still have to do `if elastic_index.load_from_source`.

I did fix the FactoryBot usage.

matthuhiggins:
`save_to_elasticsearch` is better than `Modle.elastic_index.add_record(record)`

ebarendt:
For completeness - #89

lib/elastic_record/persistence_override.rb
```diff
@@ -0,0 +1,7 @@
+module ElasticRecord
+  module PersistenceOverride
+    def save(*)
+      true
+    end
```
matthuhiggins: On closer inspection, it looks like overriding `_create_or_update` is the best choice. Overriding `save` is wrong because it blows away validation logic.

malept: I saw that when I was investigating this feature. My concern about it is that it ties into implementation details for ActiveRecord and so we run the risk of being tied to minor Rails versions, like Superstore.

matthuhiggins: Yup, which is ugly.

PR #89, opened 2018-12-21
Helper method for saving to ES

ebarendt:


matthuhiggins:
Would this call callbacks, or is this used within the callback?

ebarendt:
I thought about calling it from [here](https://github.com/data-axle/elastic_record/blob/e06b3b6e35d68172133c9b6f1835500159376d1b/lib/elastic_record/callbacks.rb#L16).

matthuhiggins:
That makes sense. I think we also need one that calls callbacks, which is out of scope here.

For that reason, a different name such as `index_to_elasticsearch` is more clear, and leaves room for a save that performs callbacks.

ebarendt:
Updated.

PR #90, opened 2019-01-17
Move test under test dir

austenmadden:
I was reading through the gem and found this test that was under lib. Nothing crazy, but thought it'd be an easy win 😃


ebarendt:
Thanks!

PR #92, opened 2019-01-29
Use instantiate

matthuhiggins:
Use klass.instantiate to load persisted records. Use klass.new results in `new_record? == true`, `persisted? == false` and ActiveModel::Dirty attributes.

### DIFF
--- a/lib/elastic_record/search_hits.rb
+++ b/lib/elastic_record/search_hits.rb
@@ -19,21 +19,10 @@ module ElasticRecord
 
     def to_records
       if model.elastic_index.load_from_source
-        hits.map { |hit| load_from_hit(hit) }
+        hits.map { |hit| model.instantiate(hit) }
       else
         model.find to_ids
       end
     end
-
-    private
-
-      def load_from_hit(hit)
-        model.new.tap do |record|
-          record.id = hit['_id']
-          hit['_source'].each do |k, v|
-            record.send("#{k}=", v) if record.respond_to?("#{k}=")
-          end
-        end
-      end
   end
 end
--- a/lib/elastic_record/search_hits.rb
+++ b/lib/elastic_record/search_hits.rb
@@ -19,9 +19,7 @@ module ElasticRecord
 
     def to_records
       if model.elastic_index.load_from_source
-        hits.map do |hit|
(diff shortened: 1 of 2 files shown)


lib/elastic_record/search_hits.rb
```diff
@@ -19,7 +19,9 @@ def to_ids
 
     def to_records
       if model.elastic_index.load_from_source
-        hits.map { |hit| load_from_hit(hit) }
+        hits.map do |hit|
+          model.respond_to?(:instantiate) ? model.instantiate(hit) : load_from_hit(hit)
+        end
```
bullfight: Here the `hit` is in the form:

```ruby
Person.instantiate('_id' => 1234, '_source' => { 'first_name' => "foo" })
```

[`Superstore::Persistence`](https://github.com/data-axle/superstore/blob/master/lib/superstore/persistence.rb) expects the hash to be in the form:

```ruby
Person.instantiate('id' => 1234, 'document' => { 'first_name' => "foo" })
```

https://github.com/data-axle/superstore/blob/6208cb519ebb5de44fdc3db4ad14ea7e8b5cc977/lib/superstore/persistence.rb#L22-L27

kunruh9: I just noticed this as well. But this change can't be for our specific needs and knows nothing about superstore.

kunruh9: Looking into leveraging this change elsewhere.

matthuhiggins: Right, let's fix instantiate then

matthuhiggins:
@bullfight See https://github.com/data-axle/superstore/pull/39

lib/elastic_record/search_hits.rb
```diff
@@ -19,7 +19,9 @@ def to_ids
 
     def to_records
       if model.elastic_index.load_from_source
-        hits.map { |hit| load_from_hit(hit) }
+        hits.map do |hit|
+          model.respond_to?(:instantiate) ? model.instantiate(hit['_source'].update('id' => hit['_id'])) : load_from_hit(hit)
+        end
```
bullfight: <bikeshed> `instantiate_from_hit(hit)` </bikeshed>

matthuhiggins:
I like bike shedding

PR #93, opened 2019-01-31
Adds ElasticRecord::FromSearchHit#from_search_hit

kunruh9:
### Changes

Adds `ElasticRecord::FromSearchHit#from_search_hit` Module.

`from_search_hit` takes a search hit and index mapping properties and deserialized the elastic search hit into valid ruby objects.

### Example

ElasticSearch returns ranges in a Ruby-unfriendly format. This PR introduces changes to convert values for range datatypes into `Range` instances. This removes the need for users of this gem to handle conversion themselves.

#### Previous Range Value Format
```ruby
range = { 'gte' => 0, 'lte' => 9 }
range.class
# => Hash
```

#### New Range Value Format
```ruby
range = 0..9
range.class
# => Range
```

### Performance
Adding deserialization with `from_search_hit` increases computation time by around  ~18%. However moving initialization from `model.new` to `model.instantiate` on https://github.com/data-axle/elastic_record/pull/92 massively reduced computation time by around 300%. This additional compute time is relatively small by comparison.

### DIFF
--- a/lib/elastic_record/search_hit.rb
+++ b/lib/elastic_record/search_hit.rb
@@ -1,7 +1,10 @@
 module ElasticRecord
   module SearchHit
-    def instantiate_from_hit(hit)
-      attrs = parse_hit hit
+
+    def from_search_hit(hit, mapping_properties = elastic_index.mapping[:properties])
+      hit = hit['_source'].merge('id' => hit['_id'])
+
+      attrs = value_from_search_hit_object(hit, mapping_properties)
 
       if respond_to?(:instantiate)
         instantiate(attrs)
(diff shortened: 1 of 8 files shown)


lib/elastic_record/search_hit.rb
```diff
@@ -0,0 +1,48 @@
+module ElasticRecord
+  module SearchHit
+    def instantiate_from_hit(hit)
```
bullfight:
```suggestion
    def from_search_hit(hit)
```

bullfight: Follow the same pattern used with `AsDocument` and inject the mapping
https://github.com/data-axle/elastic_record/blob/8cb048d5fc96d42ec3bdc2766e050a0523e09df0/lib/elastic_record/as_document.rb#L3-L11

lib/elastic_record/search_hit.rb
```diff
@@ -0,0 +1,48 @@
...
+        document
+      end
+
+      def range_fields
+        @range_fields ||= begin
+          elastic_index.mapping[:properties].select do |_, definition|
+            definition[:type].to_s.end_with? 'range'
+          end
+        end
+      end
```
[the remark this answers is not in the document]
bullfight: Follow a pattern similar to that used on `AsDocument` and select the specific types you want to convert to ranges

https://github.com/data-axle/elastic_record/blob/8cb048d5fc96d42ec3bdc2766e050a0523e09df0/lib/elastic_record/as_document.rb#L24-L45

bullfight:
@kunruh9 ￼I think there is some additional work to do here:
1. We might consider re-purposing the [Widget/WidgetPart](https://github.com/data-axle/elastic_record/blob/master/test/dummy/app/models/widget.rb) dummy/model so we don't need lots of additional test setup.
2. I am curious to see if there is some logic to be shared between `AsDocument` and `FromSearchHit`.
3. We need to carefully consider performance. `as_search_document` is only called when indexing documents. `from_search_hit` will be called for every `load_from_source` search result. 

https://github.com/data-axle/elastic_record/blob/c64d1f620e6ed8e09a0139e4ae4a1b813cd235ae/lib/elastic_record/index/documents.rb#L55-L62

kunruh9:
@bullfight Okay, those are good points. I'll work on that and see if I can get some benchmarks as well to compare performance changes.

bullfight:
@ebarendt Do you want to take a look at this before we merge?

ebarendt:
This looks good. Does this make the models loaded from source behave more like AR models?

kunruh9:
> Does this make the models loaded from source behave more like AR models?

This is a step in that direction, yes.

bullfight:
> Does this make the models loaded from source behave more like AR models?

Yes, it allows the instance to be initialized with [`instantiate`](https://github.com/data-axle/superstore/blob/11cd5c0df02bee0542c004dac02e23146c4e4c9a/lib/superstore/persistence.rb#L22-L39).

matthuhiggins:
This has a few problems:

* elastic_index.mapping[:properties] can be different than what's in the index, especially when reindexing
* elastic_index.mapping[:properties] is not a great API. Time should be put into something better than a `Hash`.
* Many fields are dynamically mapped. Relying on the ruby structure versus `get_mapping` is unreliable and does not help open source users.

It seems like we can simply rely on existence of `gte` or `lte` to determine what to do.

PR #94, opened 2019-02-14
Deserialize range types based on presence of 'gte' in hash

bullfight:
Follow up to https://github.com/data-axle/elastic_record/pull/93

**Problem**
* `elastic_index.mapping[:properties]` can be different than what's in the index, especially when reindexing
* `elastic_index.mapping[:properties]` is not a great API. Time should be put into something better than a `Hash`.
* Many fields are dynamically mapped. Relying on the ruby structure versus `get_mapping` is unreliable and does not help open source users.

**Solution**
* Change from_search_hit to de-serialize ranges by naively looking for keys `gte`
* When `'gte'` is a string de-serialize range as `date_range`
* When `'gte'` is a integer de-serialize range as `integer_range`

### DIFF
--- a/lib/elastic_record/from_search_hit.rb
+++ b/lib/elastic_record/from_search_hit.rb
@@ -26,30 +26,27 @@ module ElasticRecord
           case value
           when Hash
             hit[field] = value_from_search_hit(value)
-          when Array
-            hit[field] = value.map { |el| value_from_search_hit_object(el) }
+          when Array # type: 'nested'
+            value.each do |element|
+              if element.is_a? Hash
+                value_from_search_hit_object(element)
+              end
+            end
           end
         end
 
         hit
       end
 
-      RANGE_KEYS = %w(gte lte)
       def value_from_search_hit(value)
-        if (RANGE_KEYS & value.keys).any?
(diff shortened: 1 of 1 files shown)


lib/elastic_record/from_search_hit.rb
```diff
@@ -19,40 +19,33 @@ def from_search_hit(hit, mapping_properties = elastic_index.mapping[:properties]
...
-          when :date_range
-            value_for_date_range(value)
-          else
-            value
-          end
-
-        if value.present? || value == false
-          value
+      def value_from_search_hit(value)
+        case value['gte'] # the gte lower bound is never nil
```
dcromer: Why is this happening?

whereisciao: `value['gte']` was the most reliable value used to determine the correct handler. The previous approach was to check for both `gte` and `lte` keys in the value hash, but it triggered an unnecessary array.

bullfight: We need to infer the type of the key `lte`
* String type - `date_range`
* Integer type - `integer_range`

PR #95, opened 2019-03-15
Index suffix

ebarendt:
It can be used to append "test" to the indexes and aliases so that one
ElasticSearch server can be used for tests and development.

### DIFF
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -50,11 +50,11 @@ module ElasticRecord
     end
 
     def alias_name=(name)
-      @alias_name = name
+      @alias_name = add_suffix(name)
     end
 
     def alias_name
-      @alias_name ||= model.base_class.name.demodulize.underscore.pluralize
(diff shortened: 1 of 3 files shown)


test/elastic_record/index_test.rb
```diff
@@ -8,7 +8,10 @@ def test_copy
   end
 
   def test_alias_name
-    assert_equal 'widgets', index.alias_name
```
AlexMooney: Keep this assertion to show that setting an index_suffix causes the name to change.

lib/elastic_record/index.rb
```diff
@@ -96,6 +96,15 @@ def get(end_path, json = nil)
 
     private
 
+      def add_suffix(name)
+        suffix = ElasticRecord::Config.index_suffix
+        if suffix && !name.end_with?(suffix)
```
matthuhiggins: What is the `end_with?` check for?

ebarendt: `add_suffix` is called from `alias_name` and `alias_name=`, so I didn't want to end up with "_test_test", etc, added.

PR #96, opened 2019-03-26
Add find_hits_in_batches

malept:
Adds a convenience method that yields `SearchHits` when scrolling.

### DIFF
--- a/test/elastic_record/relation/batches_test.rb
+++ b/test/elastic_record/relation/batches_test.rb
@@ -21,8 +21,7 @@ class ElasticRecord::Relation::BatchesTest < MiniTest::Test
     Widget.elastic_relation.find_hits_in_batches do |hits|
       results << hits
     end
-    expected_ids = [[@red_widget, @blue_widget, @green_widget].to_set]
-    assert_equal expected_ids, results.map(&:to_records).map(&:to_set)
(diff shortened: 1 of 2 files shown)


AlexMooney:
Looks good.  Add a test in https://github.com/data-axle/elastic_record/blob/master/test/elastic_record/relation/batches_test.rb ?

malept:
I thought about that, but it's basically tested via `find_in_batches` and `find_ids_in_batches`.

AlexMooney:
But `find_hits_in_batches` is a new public method that content system is using.  I'm sure it works because of those tests, but if someone deletes it and the tests still pass, we could be in a bad way.

test/elastic_record/relation/batches_test.rb
```diff
@@ -16,6 +16,15 @@ def setup
   #   assert_equal [@red_widget, @blue_widget, @green_widget].to_set, results.to_set
   # end
 
+  def test_find_hits_in_batches
+    results = []
+    Widget.elastic_relation.find_hits_in_batches do |hits|
+      results << hits
+    end
+    expected_ids = [[@red_widget, @blue_widget, @green_widget].to_set]
```
kunruh9: These are `SearchHit` objects, right? Not ids.
```suggestion
    expected_hits = [[@red_widget, @blue_widget, @green_widget].to_set]
```

PR #97, opened 2019-04-17
Clear scroll after #each_slice

austenmadden:
After a batch operation (which largely use #each_slice) we should clear
the scroll by default to reduce the cost of these operations.

This implementation clears them all in 1 DELETE after yielding search results. You could probably delete as you go as well. More requests, but also cleans up sooner. I think this works, but I could see the argument for implementing that as well. Open for discussion.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -7,7 +7,7 @@ module ElasticRecord
       def initialize(elastic_index, search: nil, scroll_id: nil, keep_alive:, batch_size:)
         @elastic_index = elastic_index
         @search        = search
-        @scroll_ids    = []
+        @scroll_ids    = [scroll_id].compact
         @keep_alive    = keep_alive
         @batch_size    = batch_size
 
@@ -30,7 +30,7 @@ module ElasticRecord
       end
 
       def request_next_scroll
-        response = scroll_id.any? ? @elastic_index.scroll(scroll_ids.last, keep_alive) : initial_search_response
+        response = scroll_ids.any? ? @elastic_index.scroll(scroll_ids.last, keep_alive) : initial_search_response
         @scroll_ids << response['_scroll_id']
         response
       end
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -3,11 +3,11 @@ require 'active_support/core_ext/object/to_query'
 module ElasticRecord
   class Index
     class ScrollEnumerator
-      attr_reader :keep_alive, :batch_size, :scroll_ids
+      attr_reader :keep_alive, :batch_size, :scroll_id
       def initialize(elastic_index, search: nil, scroll_id: nil, keep_alive:, batch_size:)
         @elastic_index = elastic_index
         @search        = search
-        @scroll_ids    = [scroll_id].compact
+        @scroll_id     = scroll_id
         @keep_alive    = keep_alive
         @batch_size    = batch_size
 
@@ -17,8 +17,6 @@ module ElasticRecord
         while (hits = request_more_hits.hits).any?
           hits.each_slice(batch_size, &block)
         end
-
-        clear if scroll_ids.any?
       end
 
       def request_more_ids
@@ -30,21 +28,20 @@ module ElasticRecord
       end
 
       def request_next_scroll
-        response = scroll_ids.any? ? @elastic_index.scroll(scroll_ids.last, keep_alive) : initial_search_response
+        response = scroll_id.present? ? @elastic_index.scroll(scroll_id, keep_alive) : initial_search_response
+        new_scroll_id = response['_scroll_id']
 
-        if new_scroll_id = response['_scroll_id']
-          @scroll_ids << new_scroll_id
+        if scroll_id && new_scroll_id != scroll_id
+          @elastic_index.delete_scroll(scroll_id)
         end
+
+        @scroll_id = new_scroll_id
 
         response
       end
 
       def total_hits
         initial_search_response['hits']['total']
-      end
-
-      def clear
-        @elastic_index.clear_scroll(scroll_ids)
       end
 
       def initial_search_response
@@ -173,7 +170,7 @@ module ElasticRecord
         end
       end
 
-      def clear_scroll(scroll_ids)
+      def delete_scroll(scroll_ids)
         connection.json_delete('/_search/scroll', { scroll_id: scroll_ids })
       end
 
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -3,21 +3,22 @@ require 'active_support/core_ext/object/to_query'
 module ElasticRecord
   class Index
     class ScrollEnumerator
-      attr_reader :scroll_id, :keep_alive, :batch_size
+      attr_reader :scroll_id, :keep_alive, :batch_size, :scroll_ids
       def initialize(elastic_index, search: nil, scroll_id: nil, keep_alive:, batch_size:)
         @elastic_index = elastic_index
         @search        = search
-        @scroll_id     = scroll_id
+        @scroll_ids    = []
         @keep_alive    = keep_alive
         @batch_size    = batch_size
+
       end
 
-      def each_slice(clear_scroll_after = true, &block)
+      def each_slice(&block)
         while (hits = request_more_hits.hits).any?
           hits.each_slice(batch_size, &block)
         end
 
-        clear if clear_scroll_after
+        clear if scroll_ids.any?
       end
 
       def request_more_ids
@@ -29,13 +30,8 @@ module ElasticRecord
       end
 
       def request_next_scroll
-        if scroll_id.nil?
-          response = initial_search_response
-        else
-          response = @elastic_index.scroll(scroll_id, keep_alive)
-        end
-
-        @scroll_id = response['_scroll_id']
+        response = scroll_id.any? ? @elastic_index.scroll(scroll_ids.last, keep_alive) : initial_search_response
+        @scroll_ids << response['_scroll_id']
         response
       end
 
@@ -44,7 +40,7 @@ module ElasticRecord
       end
 
       def clear
-        @elastic_index.clear_scroll(scroll_id)
+        @elastic_index.clear_scroll(scroll_ids)
       end
 
       def initial_search_response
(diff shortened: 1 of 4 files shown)


lib/elastic_record/index/documents.rb
```diff
@@ -27,20 +30,19 @@ def request_more_hits
...
 
       def request_next_scroll
-        if scroll_id.nil?
-          response = initial_search_response
-        else
-          response = @elastic_index.scroll(scroll_id, keep_alive)
-        end
-
-        @scroll_id = response['_scroll_id']
+        response = scroll_ids.any? ? @elastic_index.scroll(scroll_ids.last, keep_alive) : initial_search_response
```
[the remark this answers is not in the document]
austenmadden: scroll_id became scroll_ids which should always be exist because of the initialize line. It is kind of confusing because I kept the signature as is, but that's mostly because it would only ever make sense in the current usage to pass a sing scroll id. (At least how it's used now)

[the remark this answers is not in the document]
AlexMooney: I see.  In the line below this one, is it possible `response['_scroll_id']` will be `nil` or non-unique?  Might need to re-compact the scroll_ids if so.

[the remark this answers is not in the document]
austenmadden: I don't think it would ever be nil (Barring a error or something, which will raise an error, so doesn't really matter), but not unique certainly according the ES docs. But that's ok, because the DELETE endpoint is very permissive. You can send it duplicates/expired scroll ids, etc and it just tells you how many it actually cleared. We just want to make sure the last result is always the end of that array for the next usage.

austenmadden:
Was there a reason that wasn't made the default to begin with?

matthuhiggins:
Yes, there was a good reason.

https://www.elastic.co/guide/en/elasticsearch/reference/6.3/search-request-scroll.html#scroll-search-context

austenmadden:
@matthuhiggins To make sure were on the same page: the way this seems to work, is to make a single scroll request without a scroll param once your done. So:

E.g
```
GET /whatever/_search?scroll=1m
POST /_search/scroll 
{
    "scroll" : "1m", 
    "scroll_id" : "DXF1ZXJ5QW5kRmV0Y2gBAAAAAAAAAD4WYm9laVYtZndUQlNsdDcwakFMNjU1QQ==" 
}
....
POST /_search/scroll 
{
    "scroll_id" : "the next id or whatever" 
}
```

The last one would free the context. That's how it seems to work in all my testing.

So the solution in this case would be to 

```ruby
      def each_slice(&block)
        while (hits = request_more_hits.hits).any?
          hits.each_slice(batch_size, &block)
        end

        @elastic_index.scroll(scroll_id, nil)
        nil
      end
```

(method always returned nil previously and the tests get mad if returns something)

austenmadden:
Or I dunno, getting some weird results out of this. Seems like it doesn't work? It's not free period, until you explicitly free it (ignoring timeout).

austenmadden:
Ok, playing around with this more: Part 50 and really getting deep into this with Kibana: A request without a keep alive will cleanup after itself: per the documentation. However, `GET /_nodes/stats/indices/search` will still reflect open contexts. I'm unsure exactly when/how it refreshes itself. Might be tied to the original search/keep_alive. Whose to say 🤷‍♂️

matthuhiggins:
The keep alive is refreshed on each request.

lib/elastic_record/index/documents.rb
```diff
@@ -167,6 +173,10 @@ def scroll(scroll_id, scroll_keep_alive)
         end
       end
 
+      def clear_scroll(scroll_ids)
+        connection.json_delete('/_search/scroll', { scroll_id: scroll_ids })
+      end
```
matthuhiggins: Belongs with other search methods. 

Name it delete_scroll

lib/elastic_record/index/documents.rb
```diff
@@ -27,20 +30,23 @@ def request_more_hits
...
       def request_next_scroll
-        if scroll_id.nil?
-          response = initial_search_response
-        else
-          response = @elastic_index.scroll(scroll_id, keep_alive)
+        response = scroll_ids.any? ? @elastic_index.scroll(scroll_ids.last, keep_alive) : initial_search_response
+
+        if new_scroll_id = response['_scroll_id']
+          @scroll_ids << new_scroll_id
         end
```
matthuhiggins: I think a scroll is returned on every request, even if the same as previous. You may need to use |=

austenmadden:
@matthuhiggins Given our discussion yesterday, how do you feel about this solution? I updated it to capture reducing the keep alive default as well.

lib/elastic_record/index/documents.rb
```diff
@@ -27,20 +30,23 @@ def request_more_hits
...
         response
       end
 
       def total_hits
         initial_search_response['hits']['total']
       end
 
+      def clear
+        @elastic_index.clear_scroll(scroll_ids)
+      end
```
matthuhiggins: Unclear method name (like my pun?)

lib/elastic_record/index/documents.rb
austenmadden: All the methods around this are making these types of operations (e.g. making a request) through the connection. Where would you like this to go and why not here?

Are you just saying you'd like to see #scroll and #delete_scroll moved above build_scroll_enumerator?

lib/elastic_record/index/documents.rb
austenmadden: A keep_alive-less scroll request won't return one (since it's technically ending the context). At the time I was experimenting with them, but I can put this back how it was.

lib/elastic_record/index/documents.rb
matthuhiggins: Ah sorry. I thought we had a Search module, like Manage. Never mind.

lib/elastic_record/index/documents.rb
matthuhiggins: Sure but you aren’t using a keep-aliveless request anymore. I thought that was a failed idea.

austenmadden: Yep, not disagreeing, just stating why it was like that.

austenmadden:
@matthuhiggins 

> This implementation clears them all in 1 DELETE after yielding search results. You could probably delete as you go as well. More requests, but also cleans up sooner. I think this works, but I could see the argument for implementing that as well. Open for discussion.

That was part of the original discussion. I'm open to it, an early implementation worked like that. I think it caused some tests to start failing so I switched to this approach, but that's likely because I wasn't account for the same scroll_id being returned (thus clearing itself).

matthuhiggins:
Right. You should deploy to beta and run some scans on the console to see how times look with the delete operation. This whole idea could be bust if you are seeing 1 second http requests

austenmadden:
@matthuhiggins Seems better to delete as we go. It's actually a rare-ish event that is switches scroll_ids at least in my testing (beta). This way it only makes the request if it needs to, and won't make a huge request at the end.

When it does make a request, it seemed to be 3 to 4ms. 15ms was the absolute worse I saw.

test/elastic_record/index/documents_test.rb
```diff
@@ -99,12 +99,36 @@ def test_expired_scroll_error
...
+  def test_delete_scroll
+    scroll_enumerator = index.build_scroll_enumerator(search: {'query' => {query_string: {query: ''}}})
+    scroll_enumerator.initial_search_response
+
+    index.delete_scroll(scroll_enumerator.scroll_id)
+    delete = index.connection.deferred_actions.first
+
+    assert_equal :json_delete, delete.method
+    assert_equal ['/_search/scroll', {scroll_id: scroll_enumerator.scroll_id}], delete.args
+  end
```
matthuhiggins: Too much internal testing here...

lib/elastic_record/index/documents.rb
```diff
@@ -27,13 +29,15 @@ def request_more_hits
       end
 
       def request_next_scroll
-        if scroll_id.nil?
-          response = initial_search_response
-        else
-          response = @elastic_index.scroll(scroll_id, keep_alive)
+        response = scroll_id.present? ? @elastic_index.scroll(scroll_id, keep_alive) : initial_search_response
```
matthuhiggins: The method checks i scroll_id.nil? two times.

test/elastic_record/index/documents_test.rb
austenmadden: The deferred action thing makes it difficult to assert anything about the request made without doing this or mocks. Would you just prefer not to test?

austenmadden: I'll just disable using the wrapper to disable deferring and use webmock.

matthuhiggins: uh please don't use webmock

austenmadden: Ok... so like; just delete it? I don't think I can do anything else to assert any behavior.

austenmadden: Guess, I can either make another request and catch the error or have to trust the stats request.

matthuhiggins:
Apologies for delay here. I’m not the best person for “availability”. Eric, Dan, Mark & Patrick are also good resources for reviews on this

austenmadden:
@ebarendt just tagged you because I know you're up earlier. Feel free to delegate to Dan, Mark or Patrick.

test/elastic_record/index/documents_test.rb
```diff
@@ -99,7 +99,25 @@ def test_expired_scroll_error
     )
 
     scroll_enumerator.request_more_hits
-    index.connection.json_delete '/_search/scroll', scroll_id: scroll_enumerator.scroll_id
+    index.delete_scroll(scroll_enumerator.scroll_id)
+    assert_raises ElasticRecord::ExpiredScrollError do
+      scroll_enumerator.request_more_hits
+    end
+  end
```
matthuhiggins: The above several lines seem out of place. Are you "borrowing" the state of the test to test that delete_scroll works?

austenmadden: Without using webmock or some other mock/spy it's difficult to assert that the context is closed (the state of elastic search stats request doesn't update quickly enough for the test), so this assertion/extra request is necessary

austenmadden:
Ugh, I meant to squash those commits, but don't feel like playing commit doctor with Master 🤦‍♂️

test/elastic_record/index/documents_test.rb
matthuhiggins: Yeah but there’s an explicit call to delete the scroll id

matthuhiggins: Yeah but there’s an explicit call to delete the scroll id

PR #99, opened 2019-05-29
Ensure provided nested object fields submit

austenmadden:
Currently in elastic record attempts to clear nested fields (set them
to null) are dropped. To allow this, we need to allow specifically
provided (in the active record sense) fields to be submitted.

For example, if you had a nested field/object sub_widget on a widget, and tried to "clear" a field on subwidget by setting the value to nil, that update would be dropped from the update sent to ElasticSearch.

### Example

```ruby
widget = Widget.create(name: 'My Widget', sub_widget: SubWidget.new(sub_field: 'hi!'))
sub_widget = widget.sub_widget

sub_widget.sub_field = nil
widget.sub_widget = sub_widget

widget.save!
```

The resulting post wouldn't include sub_field, and this the old value would persist according to the source document in ElasticSearch.

This change corrects this problem.

### DIFF
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -4,7 +4,7 @@ module ElasticRecord
       mapping_properties.each_with_object({}) do |(field, mapping), result|
         value = value_for_elastic_search field, mapping, mapping_properties
 
-        unless value.nil?
+        if !value.nil? || provided?(field)
           result[field] = value
         end
       end
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -4,7 +4,7 @@ module ElasticRecord
       mapping_properties.each_with_object({}) do |(field, mapping), result|
         value = value_for_elastic_search field, mapping, mapping_properties
 
-        if !value.nil? || provided?(field)
+        if !value.nil? || (respond_to?(:provided?) && provided?(field))
           result[field] = value
         end
       end
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -21,7 +21,7 @@ module ElasticRecord
     end
 
     def value_for_elastic_search(field, mapping, mapping_properties, is_update = false)
-      return unless (value = try(field))
+      return if (value = try(field)).nil?
 
       case mapping[:type]&.to_sym
       when :object
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -4,48 +4,42 @@ module ElasticRecord
       mapping_properties.each_with_object({}) do |(field, mapping), result|
         value = value_for_elastic_search field, mapping, mapping_properties
 
-        if !value.nil? || (respond_to?(:provided?) && provided?(field))
+        unless value.nil?
           result[field] = value
         end
       end
     end
 
-    def as_partial_update_document
-      mapping_properties = elastic_index.mapping[:properties]
+    def as_partial_update_document(mapping_properties = elastic_index.mapping[:properties])
       changed_fields = respond_to?(:saved_changes) ? saved_changes.keys : changed
 
       changed_fields.each_with_object({}) do |field, result|
         if field_mapping = mapping_properties[field]
-          result[field] = value_for_elastic_search field, field_mapping, mapping_properties
+          result[field] = value_for_elastic_search field, field_mapping, mapping_properties, true
         end
       end
     end
 
-    def value_for_elastic_search(field, mapping, mapping_properties)
-      value = try field
-      return if value.nil?
+    def value_for_elastic_search(field, mapping, mapping_properties, is_update = false)
+      return unless (value = try(field))
 
-      value =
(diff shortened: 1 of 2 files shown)


lib/elastic_record/as_document.rb
```diff
@@ -4,7 +4,7 @@ def as_search_document(mapping_properties = elastic_index.mapping[:properties])
       mapping_properties.each_with_object({}) do |(field, mapping), result|
         value = value_for_elastic_search field, mapping, mapping_properties
 
-        unless value.nil?
+        if !value.nil? || (respond_to?(:provided?) && provided?(field))
           result[field] = value
```
dcromer: Do you understand why we have separate as_search_document and as_partial_update_document functions?

austenmadden: Presumably because you have a search/source/document representation and a need to build a submission/update. 

We can change the definition of `value_for_elastic_search_object` so it doesn't delegate to `as_search_document`, and put this behavior elsewhere.

But na, not concretely, that's the point of tagging people for review with many years of experience in a given repo.

dcromer: Partial updates is not just not-synonymous, it's a [distinct ElasticSearch feature](https://www.elastic.co/guide/en/elasticsearch/guide/current/update-doc.html) that we are opting to use.  

> We can change the definition of value_for_elastic_search_object so it doesn't delegate to as_search_document, and put this behavior elsewhere.

If you are proposing changing  as_partial_update_document to call itself recursively via value_for_elasticsearch_object, I think that is the right approach.

austenmadden: I should clarify, I was aware of the Elastic Search api/feature. It's what we were debugging/working backgrounds from when we found this. I just meant that I wasn't necessarily aware of all the semantics within this repo/gem specifically.

austenmadden:
@matthuhiggins 

> Currently in elastic record attempts to clear nested fields (set them
to null) are dropped. To allow this, we need to allow specifically
provided (in the active record sense) fields to be submitted.

For example, if you had a nested field/object sub_widget on a widget, and tried to "clear" a field on subwidget by setting the value to `nil`, that update would be dropped from the update sent to ElasticSearch.

For example

```ruby
widget = Widget.new(name: 'My Widget', sub_widget: SubWidget.new(sub_field: 'hi!'))
sub_widget = widget.sub_widget

sub_widget.sub_field = nil
widget.sub_widget = sub_widget

widget.save!
```

The resulting post wouldn't include sub_field, and this the old value would persist according to the source document in ElasticSearch.

matthuhiggins:
I got lost because `Widget` uses partial updates and this makes a change to `as_search_document`, which should not be used.

I now see that `value_for_elastic_search_object` always uses `as_search_document` when it should use `as_partial_update_document` for this case. Isn't that the right fix to make?

austenmadden:
Yep, which is what @dcromer was advocating for. I'm working on it now.

matthuhiggins:
ok i need to stop using my phone during this 8 hour meeting. I'm not reading everything thoroughly.

lib/elastic_record/as_document.rb
```diff
@@ -10,42 +10,36 @@ def as_search_document(mapping_properties = elastic_index.mapping[:properties])
       end
     end
 
-    def as_partial_update_document
-      mapping_properties = elastic_index.mapping[:properties]
+    def as_partial_update_document(mapping_properties = elastic_index.mapping[:properties])
       changed_fields = respond_to?(:saved_changes) ? saved_changes.keys : changed
```
austenmadden: So there's another part to this bug where "technically" the changes on `SubWidget` aren't saved (if backed by another DB). So the bug persists. We can fix this a couple different ways, using/merging `provided` might work. However, that would change the semantics/behavior of this code. Currently noodling on this.

matthuhiggins:
I am guessing you'll need to override `changes_applied`, calling `changes_applied` on each nested object. This would go in `axle_attributes` and be similar to `validate_nestings` in terms of loop execution.

This is just a guess. Working from memory.

lib/elastic_record/as_document.rb
```diff
@@ -4,48 +4,42 @@ def as_search_document(mapping_properties = elastic_index.mapping[:properties])
...
-        when :integer_range, :float_range, :long_range, :double_range, :date_range
-          value_for_elastic_search_range(value)
-        else
-          value
-        end
-
-      if value.present? || value == false
-        value
+    def value_for_elastic_search(field, mapping, mapping_properties, is_update = false)
+      return unless (value = try(field))
```
AlexMooney: Gah!  It took me too long to figure out this line was the problem.

AlexMooney:
@matthuhiggins https://github.com/data-axle/axle_attributes/pull/75/commits/1f89f9274065ad7ed7e8d2a5a14b44c9854728b5

lib/elastic_record/as_document.rb
austenmadden: Sorry about that, I was moving the guards around a lot at first, and forgot to put this one back to the original nil check (vs falsey).

PR #102, opened 2019-07-08
multi_search support

ebarendt:
Enables `Document.multi_search([query1, query2])`. Mirrors the `search` method above it.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -155,6 +155,16 @@ module ElasticRecord
         get url, elastic_query.update('_source' => load_from_source)
       end
 
+      def multi_search(elastic_queries)
+        url = "_msearch"
+
+        queries = elastic_queries.flat_map do |query|
+          ["{}", query.update('_source' => load_from_source).to_json]
+        end
+        queries = queries.join("\n") + "\n"
+        get url, queries
+      end
+
       def explain(id, elastic_query)
         get "_explain", elastic_query
       end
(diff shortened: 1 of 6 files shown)


lib/elastic_record/index/documents.rb
```diff
@@ -155,6 +155,16 @@ def search(elastic_query, options = {})
         get url, elastic_query.update('_source' => load_from_source)
       end
 
+      def multi_search(elastic_queries)
+        url = "_msearch"
```
matthuhiggins: No reason for the `url` variable. This is made in `search` because of additional logic that adds options.

lib/elastic_record/index/documents.rb
```diff
@@ -155,6 +155,16 @@ def search(elastic_query, options = {})
...
+
+        queries = elastic_queries.flat_map do |query|
+          ["{}", query.update('_source' => load_from_source).to_json]
```
matthuhiggins: `load_from_source` should not be here. It is not the job of this class to mutate the query and add `load_from_source`. The user of this method should pass that in.

lib/elastic_record/index/documents.rb
```diff
@@ -155,6 +155,16 @@ def search(elastic_query, options = {})
         get url, elastic_query.update('_source' => load_from_source)
       end
 
+      def multi_search(elastic_queries)
+        url = "_msearch"
+
+        queries = elastic_queries.flat_map do |query|
```
matthuhiggins: This is more like "headers_and_bodies" or "headers_and_searches". Inside each "search" is a query and aggregations.

PR #103, opened 2019-07-31
Make ElasticRecord ES 7.X Compatible

austenmadden:
ElasticRecord is in need of a fresh coat of paint! There's several issues raised by ES 7 that need resolved. Specifically what I'm attempting to resolve:

- New mapping structure (removal of mapping types) [link](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html)
- total_hits returns an object now [link](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#hits-total-now-object-search-response)

### DIFF
--- a/lib/elastic_record/index/manage.rb
+++ b/lib/elastic_record/index/manage.rb
@@ -8,10 +8,12 @@ module ElasticRecord
       end
 
       def create(index_name = new_index_name, setting_overrides: {})
-        connection.json_put "/#{index_name}", {
-          "mappings" => mapping ,
+        mapping_params = {
+          "mappings" => (custom_mapping_type_name? ? { mapping_type => mapping } : mapping),
           "settings" => settings.merge(setting_overrides)
         }
+
+        connection.json_put "/#{index_name}?include_type_name=#{custom_mapping_type_name?}", mapping_params
         index_name
       end
 
--- a/lib/elastic_record/index/manage.rb
+++ b/lib/elastic_record/index/manage.rb
@@ -13,6 +13,7 @@ module ElasticRecord
           "settings" => settings.merge(setting_overrides)
         }
 
+        # TODO: Remove include_type_name when ES8 support is added
         connection.json_put "/#{index_name}?include_type_name=#{custom_mapping_type_name?}", mapping_params
         index_name
       end
--- a/lib/elastic_record/index/manage.rb
+++ b/lib/elastic_record/index/manage.rb
@@ -9,9 +9,7 @@ module ElasticRecord
 
       def create(index_name = new_index_name, setting_overrides: {})
         connection.json_put "/#{index_name}", {
-          "mappings" => {
-            mapping_type => mapping
-          },
+          "mappings" => mapping ,
           "settings" => settings.merge(setting_overrides)
         }
         index_name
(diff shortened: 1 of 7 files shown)


lib/elastic_record/relation.rb
```diff
@@ -32,7 +32,8 @@ def becomes(klass)
     end
 
     def count
-      search_results['hits']['total']
+      count = search_results['hits']['total']
+      count.is_a?(Hash) ? count['value'] : count
```
malept: DRY this up with the one in `index/search.rb`.

lib/elastic_record/search_hits.rb
```diff
@@ -1,15 +1,16 @@
...
 
     class << self
       def from_response(response)
-        new(response['hits']['hits'])
+        new(response['hits']['hits'], response['hits']['total'])
       end
     end
 
-    def initialize(hits)
+    def initialize(hits, total = nil)
```
malept: Keyword arg?
```suggestion
    def initialize(hits, total: nil)
```

lib/elastic_record/index/manage.rb
```diff
@@ -8,12 +8,12 @@ def create_and_deploy(index_name = new_index_name)
...
-        connection.json_put "/#{index_name}", {
-          "mappings" => {
-            mapping_type => mapping
-          },
+        mapping_params = {
+          "mappings" => (custom_mapping_type_name? ? { mapping_type => mapping } : mapping),
           "settings" => settings.merge(setting_overrides)
         }
+
+        connection.json_put "/#{index_name}?include_type_name=#{custom_mapping_type_name?}", mapping_params
```
malept: Perhaps add:
```suggestion
        # TODO: Remove include_type_name when ES8 support is added
        connection.json_put "/#{index_name}?include_type_name=#{custom_mapping_type_name?}", mapping_params
```

austenmadden: How about:

`TODO: Evaluate necessity of this param during ES8 update`?

It was necessary to provide it to bypass some kind of error checking ES7 was doing, even thought the format of the mapping seemed to map their new structure + the param is default faulse. I guess I just want to keep it backwards compatible a version to eliminate update woes.

malept: My wording was because of what I read in the removal of mapping types link you had:

> **Elasticsearch 8.x**
> * The `include_type_name` parameter is removed.

austenmadden: Right, but then it's usage should be conditional vs straight up removed if we want the gem to work with both versions while we update ES.

austenmadden: eh, it's a comment not worth arguing about. We can always update it.

lib/elastic_record/search_hits.rb
```diff
@@ -1,15 +1,16 @@
...
       def from_response(response)
-        new(response['hits']['hits'])
+        new(response['hits']['hits'], total: response['hits']['total'])
       end
     end
 
-    def initialize(hits)
+    def initialize(hits, total: nil)
       @hits  = hits
+      @count = total.is_a?(Hash) ? total['value'] : total
```
malept: I think `count` would imply that it's the number of values in the `hits` accessor, whereas `total` would imply that it's the overall count. So the accessor would need to be renamed.

PR #104, opened 2019-08-06
Remove type param/routing in indexing docs/search

austenmadden:
Deprecated/gone in 7+, not needed in 6. Since Data Axle is already not using and/or reliant on this, start stripping the rest of it out.

There's still some alternative uses that mapping_type serves for has_child queries and such, so I can't completely remove all of it as I'd like, but this will at least get rid of some observed ES7 deprecation warnings.

### DIFF
--- a/lib/elastic_record/index.rb
+++ b/lib/elastic_record/index.rb
@@ -89,7 +89,7 @@ module ElasticRecord
 
     def get(end_path, json = nil)
       path = "/#{alias_name}"
-      path += "/#{mapping_type}"
+      path += "/#{mapping_type}" if custom_mapping_type_name?
       path += "/#{end_path}"
 
       connection.json_get path, json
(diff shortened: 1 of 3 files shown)


lib/elastic_record/index.rb
```diff
@@ -89,7 +89,7 @@ def real_connection
 
     def get(end_path, json = nil)
       path = "/#{alias_name}"
-      path += "/#{mapping_type}"
+      path += "/#{mapping_type}" if custom_mapping_type_name?
```
matthuhiggins: Preferably, `custom_mapping_type_name?` is deleted from the code

matthuhiggins:
I think we should try moving synonyms entirely to the app

austenmadden:
@matthuhiggins wrong PR?

matthuhiggins:
yes wrong pr

PR #105, opened 2019-08-09
Add es6 mode; use _type in bulk requests

austenmadden:
Unfortunately, I had missed a bit where the bulk api in ES6 requires type parameters hard stop. (You can provide them in the routing as well). Teaches me to not run the tests on both versions 🤦‍♂️.

~I'm not committed to merging this. elastic_record 5.6.0 can run in both versions (ES6.X & 7.X) though it will produce more deprecation warnings in 7.X.~ Feeling much better about where this is going. We can mostly get rid of mapping_type entirely just hard code '_doc' where appropriate now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ development:
 
 ## ElasticSearch 6.x Compatibility ##
 
-In order to work correctly set the config option `set es6_mode` to true.
+In order to work correctly set the config option `es6_mode` to true.
 ```ruby
 # config/initializers/elastic_search.rb
 ElasticRecord.configure do |config|
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ end
 
 ```yaml
 # config/elasticsearch.yml:
-```yaml
 development:
   servers: es1.example.com:9200
   timeout: 10
@@ -36,7 +35,7 @@ development:
 
 ## ElasticSearch 6.x Compatibility ##
 
-In order to work correctly set the config option `set es6`_mode to true.
+In order to work correctly set the config option `set es6_mode` to true.
 ```ruby
 # config/initializers/elastic_search.rb
 ElasticRecord.configure do |config|
@@ -203,9 +202,9 @@ class Product
 end
 ```
 
-Mapping types are removed in ElasticSearch 7.x. While you can still set
-mapping type most associated behaviors have since been removed. When used with
-ElasticSearch 6.x, this gem will use '_doc'.
+Mapping types are removed in ElasticSearch 7.x. While you can still set mapping type, most
+associated behaviors have since been removed. When used with ElasticSearch 6.x, this gem will use
+'_doc'.
 
 ```ruby
 class Product
--- a/README.md
+++ b/README.md
@@ -27,10 +27,21 @@ end
 
 ```yaml
 # config/elasticsearch.yml:
+```yaml
 development:
   servers: es1.example.com:9200
   timeout: 10
   retries: 2
+```
+
+## ElasticSearch 6.x Compatibility ##
+
+In order to work correctly set the config option `set es6`_mode to true.
+```ruby
+# config/initializers/elastic_search.rb
+ElasticRecord.configure do |config|
+  config.es6_mode = true
+end
 ```
 
 ## Search API ##
@@ -192,7 +203,9 @@ class Product
 end
 ```
 
-Mapping types will be removed in ElasticSearch 7.x.  To rename the default mapping type (`_doc`), use `elastic_index.mapping_type`:
+Mapping types are removed in ElasticSearch 7.x. While you can still set
+mapping type most associated behaviors have since been removed. When used with
+ElasticSearch 6.x, this gem will use '_doc'.
 
 ```ruby
 class Product
--- a/README.md
+++ b/README.md
@@ -33,16 +33,6 @@ development:
   retries: 2
 ```
 
-## ElasticSearch 6.x Compatibility ##
-
-In order to work correctly set the config option `es6_mode` to true.
-```ruby
-# config/initializers/elastic_search.rb
-ElasticRecord.configure do |config|
-  config.es6_mode = true
-end
-```
-
 ## Search API ##
 
 ElasticRecord adds the method 'elastic_search' to your models. It works similar to active_record scoping:
--- a/README.md
+++ b/README.md
@@ -192,18 +192,6 @@ class Product
 end
 ```
 
-Mapping types are removed in ElasticSearch 7.x. While you can still set mapping type, most
-associated behaviors have since been removed. When used with ElasticSearch 6.x, this gem will use
-'_doc'.
-
-```ruby
-class Product
-  include ElasticRecord::Model
-
-  elastic_index.mapping_type = 'product'
-end
-```
-
 ### Inheritance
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -27,12 +27,23 @@ end
 
 ```yaml
 # config/elasticsearch.yml:
+```yaml
```
malept: This is duplicate (see two lines up)
```suggestion
```

README.md
```diff
@@ -27,12 +27,23 @@ end
...
 development:
   servers: es1.example.com:9200
   timeout: 10
   retries: 2
 ```
 
+## ElasticSearch 6.x Compatibility ##
+
+In order to work correctly set the config option `set es6`_mode to true.
```
malept:
```suggestion
In order to work correctly set the config option `set es6_mode` to `true`.
```

README.md
```diff
@@ -192,7 +203,9 @@ class Product
 end
 ```
 
-Mapping types will be removed in ElasticSearch 7.x.  To rename the default mapping type (`_doc`), use `elastic_index.mapping_type`:
+Mapping types are removed in ElasticSearch 7.x. While you can still set
+mapping type most associated behaviors have since been removed. When used with
```
malept:
```suggestion
mapping type, most associated behaviors have since been removed. When used with
```

README.md
austenmadden: Editor markdown formatting woes 🤕

austenmadden:
`bin/rails test` with main app:
- [x] ES 6.8
- [x] ES 7.1

`bundle exec rake test`:
- [x] ES 6.8
- [x] ES 7.1

- [ ] QA
- [ ] ???
- [ ] Profit

malept:
Is there a good way to automatically determine when to use "ES6 mode", as opposed to having to set a config option?

I think it would be best if you added Elasticsearch 6 to the [Travis build matrix](https://docs.travis-ci.com/user/build-matrix/).

austenmadden:
@malept There's https://www.elastic.co/guide/en/elasticsearch/reference/7.1/info-api.html#info-api

You can get the build hash/date. It could set/memoize es6_mode when used.

lib/elastic_record/version.rb
```diff
@@ -0,0 +1,11 @@
+module ElasticRecord
+  class Version
+    include Model
+
+    def self.es6?
+      @es6_mode ||= elastic_connection.json_get('/')
+        .dig('version', 'number')
+        .start_with?('6.')
```
malept: I think you want a safe navigation operator if you're going to be `.dig`ging.

```suggestion
        &.start_with?('6.')
```

test/elastic_record/index/documents_test.rb
```diff
@@ -151,9 +162,12 @@ def test_bulk_inheritance
         InheritedWidget.elastic_index.index_document '5', color: 'green'
 
         expected = [
-          {index: {_index: index.alias_name, _id: "5"}},
+          {index: {_index: index.alias_name, _id: "5", _type: '_doc'}},
```
malept: Perhaps have a helper for these expected index hashes that adds _type for Elasticsearch 6.

lib/elastic_record/version.rb
```diff
@@ -0,0 +1,11 @@
...
+    end
+  end
+end
```
malept: Creating a new class to include Model seems a little heavyweight just so you can access one method (although I appreciate the API). Maybe add a class method on Model instead?

austenmadden: Well, the idea behind the class was to make it accessible outside of the scope of a specific model/index.

malept: I was thinking `ElasticRecord::Model.es6?` but your change to the `ElasticConnection` concern is fine.

lib/elastic_record/index/documents.rb
```diff
@@ -66,6 +68,7 @@ def delete_document(id, parent: nil, index_name: alias_name)
 
         if batch = current_bulk_batch
           instructions = { _index: index_name, _id: id, retry_on_conflict: 3 }
+          instructions[:_type] = '_doc' if ElasticRecord::Version.es6?
```
matthuhiggins: Is "_doc" necessary in ES 6.8? I know that some endpoints removed the need to use _doc between ES 6.7 and ES 6.8, such as the Get Mapping API, which no longer requires _doc to work.

lib/elastic_record/index/manage.rb
```diff
@@ -9,12 +9,12 @@ def create_and_deploy(index_name = new_index_name)
...
       def create(index_name = new_index_name, setting_overrides: {})
         mapping_params = {
-          "mappings" => (custom_mapping_type_name? ? { mapping_type => mapping } : mapping),
+          "mappings" => mapping,
           "settings" => settings.merge(setting_overrides)
         }
 
         # TODO: Remove include_type_name when ES8 support is added
-        connection.json_put "/#{index_name}?include_type_name=#{custom_mapping_type_name?}", mapping_params
+        connection.json_put "/#{index_name}?include_type_name=false", mapping_params
```
matthuhiggins: The current code works on both ES 6.8 and ES 7. What's the need for this?

matthuhiggins:
This is also wrong for re-indexing. @pumpkineater Are we going to reindex our 6.8 index first with new standards before migrating to ES7?

pumpkineater:
It's probably best to reindex the older ones anyway. And places needs index.query.default_field set.

README.md
```diff
@@ -192,7 +192,9 @@ class Product
 end
 ```
 
-Mapping types will be removed in ElasticSearch 7.x.  To rename the default mapping type (`_doc`), use `elastic_index.mapping_type`:
+Mapping types are removed in ElasticSearch 7.x. While you can still set mapping type, most
+associated behaviors have since been removed. When used with ElasticSearch 6.x, this gem will use
+'_doc'.
```
matthuhiggins: Delete this whole section.

test/elastic_record/index/documents_test.rb
```diff
@@ -151,16 +152,22 @@ def test_bulk_inheritance
...
+
+
         assert_equal expected, index.current_bulk_batch
       end
     end
   end
 
   private
 
+    def add_version_params(hash)
```
matthuhiggins: add_es6_params

lib/elastic_record/version.rb
```diff
@@ -0,0 +1,11 @@
+module ElasticRecord
+  class Version
+    include ElasticConnection
+
+    def self.es6?
+      @es6 ||= elastic_connection.json_get('/')
+        .dig('version', 'number')
+        &.start_with?('6.')
```
matthuhiggins: Does ES 7.2 crash when _doc is provided? Curious what will happen during the rolling upgrade from 6.8 to 7.2

austenmadden: It throws deprecation warnings and I think will eventually error in 8 (but I'm not certain).

austenmadden:
@matthuhiggins To your question re: version, I wouldn't expect anything to crash going from ES6 => ES7.

While bulk requests without _type breaking 6.8, the reverse isn't true. Just throws warnings (for now).

PR #106, opened 2019-08-13
Ensure relations return full total of hits

austenmadden:
ElasticSearch introduced a new feature where a search has the total
limited to 10k by default for all requests. While this might make sense
for some usages, elastic_record is designed to be an ORM. You wouldn't
expect a SQL COUNT to lie to you unless you gave it a magic word.

### DIFF
--- a/lib/elastic_record/relation/hits.rb
+++ b/lib/elastic_record/relation/hits.rb
@@ -13,7 +13,7 @@ module ElasticRecord
 
       def search_results
         @search_results ||= begin
-          options = { typed_keys: true }
+          options = { typed_keys: true, track_total_hits: true }
           options[:search_type] = search_type_value if search_type_value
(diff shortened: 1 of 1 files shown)


lib/elastic_record/relation/hits.rb
```diff
@@ -13,7 +13,7 @@ def search_hits
 
       def search_results
         @search_results ||= begin
-          options = { typed_keys: true }
+          options = { typed_keys: true, track_total_hits: true }
```
ebarendt, line 16: @matthuhiggins this causes the optimization you pointed me to to not be hit for nearly everything we do.

https://www.elastic.co/blog/faster-retrieval-of-top-hits-in-elasticsearch-with-block-max-wand

Matching does not hit this, however, because we do _msearch which needs to be parsed differently.

PR #108, opened 2019-08-21
Remove version

austenmadden:
It's ES7 only now.


ebarendt:
don't forget the travis config?

austenmadden:
Ah good, point 🤦‍♂️

PR #110, opened 2019-09-11
Use new update api path

austenmadden:
Elasticsearch changed this path in 7 to /{index}/_update/{id}

Resolves the deprecation warning:
```
[types removal] Specifying types in document update requests is deprecated, use the endpoint /{index}/_update/{id} instead.
```

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -53,7 +53,7 @@ module ElasticRecord
           batch << { update: instructions }
           batch << params
         else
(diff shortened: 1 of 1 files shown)


matthuhiggins:
Nice. I wonder if other instances of _doc will go away in 8.0

matthuhiggins:
This is the other one I'm thinking of. I'd be surprised if they leave _doc around forever here: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-get.html

I guess there is the _source/:id path. They could change that to /widget/id/_source instead.

PR #111, opened 2019-09-11
Ensure that nested objects are fully indexed when a root document is partially updated

dcromer:
**Problem**

#107 introduced a regression where inner object fields being set to nil were not being indexed as null.

**Solution**

Ensure all mapped fields are always specified for every inner object, whether the root document is being wholly- or partially-indexed.

### DIFF
--- a/test/elastic_record/as_document_test.rb
+++ b/test/elastic_record/as_document_test.rb
@@ -30,47 +30,52 @@ class ElasticRecord::AsDocumentTest < MiniTest::Test
 
   class SpecialFieldsModel
     include TestModel
-    attr_accessor :meta, :book_length
+    attr_accessor :author, :book_length, :commenters, :meta,
 
     class Author
       include TestModel
       attr_accessor :name, :salary_estimate
+
+      def self.mapping_properties
+        {
+          name: { type: :string },
+          salary_estimate: { type: :integer_range }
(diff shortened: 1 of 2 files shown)


test/elastic_record/as_document_test.rb
```diff
@@ -30,47 +30,52 @@ def test_as_partial_update_document
...
+    record.commenters = [
+      SpecialFieldsModel::Author.new(name: 'Jonny'),
+      SpecialFieldsModel::Author.new(name: 'Jonny')
+    ]
 
     doc = record.as_search_document
 
     assert_equal({name: 'Jonny', salary_estimate: { 'gte' => 250, 'lte' => nil }}, doc[:author])
-    assert_equal([{name: 'Jonny'}, {name: 'Jonny'}], doc[:commenters])
+    expected_commenter = { name: 'Jonny', salary_estimate: nil }
```
dcromer: This is the important assertion here.  If we want to use the same logic for wholly- and partially-updating inner objects, then _all_ inner object fields will _always_ be serialized.

matthuhiggins: This test / behavior is actually wrong. No need to perform nil updates to nested types. Only to need to do this for objects, where partial indexing is supported.

PR #112, opened 2019-09-17
Fix nested object indexing

dcromer:
**Problem**

#111 accidentally introduced behavior where nested definitions were also being fully index when the parent document was being partially updated.

**Solution**

Ensure that only objects are fully indexed during a partial update, not nested types.

### DIFF
--- a/test/elastic_record/as_document_test.rb
+++ b/test/elastic_record/as_document_test.rb
@@ -30,31 +30,31 @@ class ElasticRecord::AsDocumentTest < MiniTest::Test
 
   class SpecialFieldsModel
     include TestModel
-    attr_accessor :author, :book_length, :commenters, :meta,
+    define_attributes :author, :book_length, :commenters, :meta
 
     class Author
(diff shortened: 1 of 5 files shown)


test/elastic_record/as_document_test.rb
```diff
@@ -30,31 +30,31 @@ def test_as_partial_update_document
 
   class SpecialFieldsModel
     include TestModel
-    attr_accessor :author, :book_length, :commenters, :meta,
+    define_attributes :author, :book_length, :commenters, :meta
 
     class Author
       include TestModel
-      attr_accessor :name, :salary_estimate
+      define_attributes :name, :salary_estimate
```
dcromer: SpecialFieldsModel wasn't correctly implemented, so dirty tracking wasn't working.

matthuhiggins: A quick review tells me we could cleanup the models used for testing.

PR #116, opened 2019-12-27
Use ruby 2.6 boundless ranges

matthinea:
[Basecamp TODO](https://3.basecamp.com/4119850/buckets/9653296/todos/2293987115)

### Problem

Elastic Record checks for ranges with Float::INFINITY as their upper bounds. However, content system no longer uses ranges with Float::INFINITY as the upper bound.

### Solution

Remove handlers for ranges with Float::INFINITY as the upper bound.

### Status

This is in use by content system on QA3.

### DIFF
--- a/lib/elastic_record/as_document.rb
+++ b/lib/elastic_record/as_document.rb
@@ -45,7 +45,7 @@ module ElasticRecord
 
     def value_for_elastic_search_range(range)
       gte = range.begin unless range.begin == -Float::INFINITY
-      lte = range.end unless range.end == Float::INFINITY
+      lte = range.end
 
       {'gte' => gte, 'lte' => lte}
     end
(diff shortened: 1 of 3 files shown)


matthinea:
I think our Travis instances need to be upgraded to Ruby 2.6.

j1wilmot:
Yep, travis is failing because it's using ruby 2.5.3 which doesn't have endless ranges. This is set locally in the `.travis.yml` file. You could even bump past 2.6 to use 2.7.0 which was released on Dec 25th. http://rubies.travis-ci.org/

matthinea:
Going to merge https://github.com/data-axle/content_system/pull/4177 first and then thoroughly test https://github.com/data-axle/content_system/pull/4197 before merging this in.

PR #117, opened 2020-01-08
Hackable Relation

bullfight:
**Problem**
Modifying the behavior of query building on Relation is difficult because the methods that build the elastic record search pass in variables instead of calling them directly on the relation instance.

**Solution**
Change the method signature of `build_query_filter` to reference `filter_values` and `query_value` directly on the instance instead of passing references via variables.

### DIFF
--- a/lib/elastic_record/relation/search_methods.rb
+++ b/lib/elastic_record/relation/search_methods.rb
@@ -193,8 +193,8 @@ module ElasticRecord
         end
 
         def build_query_and_filter
-          query = build_query
-          filter = build_filter
+          query = build_search_query
+          filter = build_search_filter
 
           query_and_filter = if filter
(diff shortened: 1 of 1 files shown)


lib/elastic_record/relation/search_methods.rb
```diff
@@ -207,12 +207,16 @@ def build_query_and_filter(query, filters)
...
+          if query_value.is_a?(String)
+            Arelastic::Queries::QueryString.new query_value
+          else
+            query_value
           end
+        end
 
-          query
+        def build_search_filter
+          build_filter(filter_values)
```
kunruh9: Why not do the same thing here? `build_filter(filters)` can be directly replaced by `build_search_filter`. The old method doesn't seem to be used anywhere else.

bullfight:
I am closing this in favor of another approach.

PR #119, opened 2020-03-04
Fix Thread Safety Problem in Bulk Batching

johnkeith:
See [Basecamp discussion](https://3.basecamp.com/4119850/buckets/9653296/messages/2466524698) for details.

### DIFF
--- a/lib/elastic_record/connection.rb
+++ b/lib/elastic_record/connection.rb
@@ -5,7 +5,7 @@ module ElasticRecord
     attr_accessor :servers, :options
     attr_accessor :request_count, :current_server
     attr_accessor :max_request_count
-    attr_accessor :bulk_actions
+
     def initialize(servers, options = {})
       self.servers = Array(servers)
 
@@ -14,7 +14,6 @@ module ElasticRecord
       self.request_count      = 0
       self.max_request_count  = 100
       self.options            = options.symbolize_keys
-      self.bulk_actions       = {}
     end
 
     def head(path)
--- a/lib/elastic_record/connection.rb
+++ b/lib/elastic_record/connection.rb
@@ -14,7 +14,7 @@ module ElasticRecord
       self.request_count      = 0
       self.max_request_count  = 100
       self.options            = options.symbolize_keys
-      self.bulk_actions       = nil
+      self.bulk_actions       = {}
     end
 
     def head(path)
--- a/elastic_record.gemspec
+++ b/elastic_record.gemspec
@@ -2,7 +2,7 @@
 
 Gem::Specification.new do |s|
   s.name = 'elastic_record'
-  s.version = '5.6.1'
+  s.version = '5.6.2'
   s.summary = 'An Elasticsearch querying ORM'
   s.description = 'Find your records with Elasticsearch'
 
--- a/lib/elastic_record/index/deferred.rb
+++ b/lib/elastic_record/index/deferred.rb
@@ -11,11 +11,9 @@ module ElasticRecord
         attr_accessor :index
         attr_accessor :deferred_actions
         attr_accessor :writes_made
-        attr_accessor :bulk_actions
 
         def initialize(index)
           @index = index
-          @bulk_actions = nil
           reset!
         end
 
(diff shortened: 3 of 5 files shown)


lib/elastic_record/connection.rb
```diff
@@ -14,7 +14,7 @@ def initialize(servers, options = {})
       self.request_count      = 0
       self.max_request_count  = 100
       self.options            = options.symbolize_keys
-      self.bulk_actions       = nil
+      self.bulk_actions       = {}
```
ebarendt: Instead of this, try

```ruby
Thread.current['bulk_actions'] = nil
```

Look at this project for other uses of `Thread.current`

AlexMooney:
This looks very promising!  Could definitely explain ES being out of sync.

whereisciao:
> This looks very promising! Could definitely explain ES being out of sync.

What problems were you guys seeing? Was it with the DataImport code or something else? I ask so Support can keep an eye on things.

AlexMooney:
@whereisciao it's been a longstanding problem that ES can desync from postgres.  I can all but guarantee that something is out of sync in production, at any given time.  [Here's an example](https://data-axle.com/people?scope=9afc6431518&columns=person,emails,emails_count).

matthuhiggins:
BTW: https://github.com/data-axle/elastic_record/pull/59

That one was reverted due to minor gotchas.

johnkeith:
@matthuhiggins, @ebarendt - that is interesting, do y'all remember any of the gotcha scenarios?

In lieu of the problems we are seeing now, how about we roll this out and handle the corner cases as they arise?

ebarendt:
Honeybadgers probably.

What's up with the failing tests?

johnkeith:
@ebarendt - the tests are good now. The problem was in how we were reassigning that current_bulk_batch value here https://github.com/data-axle/elastic_record/pull/119/commits/e1ab73b2620be3cd69dcce57eecc0a1191b54f44#diff-52b112257e8727735e5a2989a0fb4673R112. I've made things more explicit and the tests are happy. 

I'm going to get this pushed to QA for further testing.

matthuhiggins:
Curious: are you trying to run concurrency in production as a new idea?

johnkeith:
@matthuhiggins no, this is in response to a bug we were dealing with where RecordLists were not being properly populated or truncated. This is just to try to get things working properly in Sidekiq when we have multiple workers running bulk batches.

johnkeith:
These changes are out on QA3 for testing.

matthuhiggins:
>  This is just to try to get things working properly in Sidekiq when we have multiple workers running bulk batches.

In the same thread?

ebarendt:
Alex and I have seen, in development, problems where PG counts != ES counts after doing large operations. I use the sidekiq `-c 4` option usually, so I'm fairly certain this contributes to the problem.

Production-wise, not sure if this would cause record lists being out of sync since they're all `-c 1`.

matthuhiggins:
> I use the sidekiq -c 4 option usually, so I'm fairly certain this contributes to the problem.

No question that this causes bugs, if you concurrency is greater than 1. It's good to fix this bug.

> Production-wise, not sure if this would cause record lists being out of sync since they're all -c 1.

This will not help anything in production.

johnkeith:
@matthuhiggins - @pumpkineater brought up the same point, that Production is 1:1 in terms of workers/threads. 

It makes sense that we thought we had the bug reproduced in QA, as the worker that processed the adding/truncating jobs for RecordLists has two threads. 

Adding to and truncating RecordLists in production seems to be working without issues as I am testing now, so that means the reason for the bug with that Tier 1 list must be something else. 😞

ebarendt:
In the past they have uploaded tier1/tier2 files with duplicate igids in them. Check that first.

johnkeith:
Thanks @ebarendt, I'll take a look. I'm going to merge this one in, since it still fixes an issue. Then I'll bump the gem version over on content system. I also see Scrape and Transfer station are both running Sidekiq with workers that have multiple threads - maybe we should update those as well? (I'm not sure about auth time, no access to Sidekiq).

ebarendt:
Scrape does use ES - or did at one point. Transfer Station does not.

PR #121, opened 2020-04-26
Fix `where` and add `left_outer_joins`

BenKanouse:
Hello,

Over the week I was chatting with @kstevens715 and he mentioned that this project had an issue with the `left_outer_joins` method, so I thought I would take a look at the project.

While reviewing the project I think I found several bugs. However, I found so many bugs that I am worried I miss-understood the API of this project. Here are the issues I found:

- [ ] If where is used with filter would only work if filter already removed the correct records.
- [ ] `where` doesn't work with most of the methods in Relation::FinderMethods module. This is because the `where` is only applied in the `find_hits` & `to_a` methods.
- [ ] ElasticSearch indexes aren't cleared between tests, so the tests could step on each other.
- [ ] The test database had miss-matching data types between wearhouses.id and warehouses_id
- [ ] Order is broken now *I did this on this branch*. I'm not using find and it turns out find is order specific.

### DIFF
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -54,7 +54,7 @@ module ElasticRecord
       if klass.elastic_index.load_from_source
         search_hits.hits.map { |hit| klass.from_search_hit(hit) }
       else
-        klass.find search_hits.to_ids
+        klass.where(klass.primary_key => search_hits.to_ids).to_a
       end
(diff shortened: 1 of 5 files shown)


BenKanouse:
Sorry I meant to open this PR between my branch and my copy of master while I was working on this.

Nothing to see here.

PR #124, opened 2020-04-28
Fix Thread Safety Problem in Bulk Batching - Take Two

johnkeith:
See [Basecamp discussion](https://3.basecamp.com/4119850/buckets/9653296/messages/2466524698) for background.

The gist is that we have thread safety problems before due to shared bulk actions array being accessed and cleared by multiple threads. 

To solve the problem, this change makes it so that ElasticRecord does not create a new Connection object for every Index, but instead has one shared Connection object that is used. The ConnectionHandler class handles bootstrapping this single Connection object and allows for using a DeferredConnection in the case of testing.

This has been tested with Content System and all the tests passed - see this [PR](https://github.com/data-axle/content_system/pull/4784).

This is a follow up to https://github.com/data-axle/elastic_record/pull/119.


johnkeith:
Closing for a bit while I consider how best to handle the reset! method - seems it is clearing out all indexes when running tests....

PR #125, opened 2020-05-20
Remove Arelastic link

johnkeith:
The link to this file in the Arelastic gem is broken.

@matthuhiggins - is there a new place this link should point? I was uncertain looking at the Arelastic source in `lib/arelastic/builders/`.


matthuhiggins:
https://github.com/matthuhiggins/arelastic/blob/master/lib/arelastic/builders/queries.rb

matthuhiggins:
Deleting is fine too.

PR #127, opened 2021-01-29
Add rake task to create indexes idempotently

betesh-at-data-axle:
This change will allow content_system's `bin/setup` to be idempotent, as Rails intended for it to be.

### DIFF
--- a/lib/elastic_record/tasks/index.rake
+++ b/lib/elastic_record/tasks/index.rake
@@ -32,6 +32,19 @@ namespace :index do
   desc "Recreate index for CLASS or all models."
   task reset: ['index:drop', 'index:create']
 
+  desc "Create index for all models that don't already have one"
+  task create_missing: :environment do
+    ElasticRecord::Task.get_models.each do |model|
+      index = model.elastic_index
+      if (all_names = index.all_names).empty?
+        index_name = index.create_and_deploy
+        puts "Created #{model.name} index (#{index_name})"
+      else
+        puts "#{model.name} already has an index (#{all_names.last})"
+      end
+    end
+  end
+
   task update_mapping: :environment do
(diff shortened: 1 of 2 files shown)


lib/elastic_record/tasks/index.rake
```diff
@@ -32,6 +32,19 @@ namespace :index do
...
 
+  desc "Create index for all models that don't already have one"
+  task create_missing: :environment do
+    ElasticRecord::Task.get_models.each do |model|
+      index = model.elastic_index
+      if (all_names = index.all_names).empty?
+        index_name = index.create_and_deploy
+        puts "Created #{model.name} index (#{index_name})"
+      else
+        puts "#{model.name} already has #{'index'.pluralize(all_names.size)} (#{all_names.to_sentence})"
```
AlexMooney, line 43: This output will be a huge mess for those of us who have had our development environments around for quite some time.  How about just showing the latest index name?  (Or the _deployed_ index, though I couldn't figure out how to get that after 5 minutes of hunting.  It looks like we have an alias that points to the latest index…)

```suggestion
        puts "#{model.name} already has #{'index'.pluralize(all_names.size)} (#{all_names.last})"
```

AlexMooney (approved):
Very nice!  But please don't print all of the index names for the sake of us old timers! 😆

lib/elastic_record/tasks/index.rake
betesh-at-data-axle: I tested this locally and it only ever prints one index name per model anyway.  But you're right, `.last` is better since it makes it clear what to expect.

betesh-at-data-axle: Fixed and squashed

PR #128, opened 2021-03-16
Fix method_missing usage to support Ruby 3

betesh-at-data-axle:
We already broke Ruby < 2.7 compatibility in
3fb46a08ece814405f6a3d74ce9d57975824806d, so we
also update the gemspec to reflect that.


j1wilmot (approved):
I like the move to github actions.

PR #132, opened 2021-04-28
Run elasticsearch in a docker container during CI build

betesh-at-data-axle:


AlexMooney (approved):
LGTM.  What are the default `health-interval`, `health-timeout`, and `health-retries` values?  Perhaps we could use the defaults instead of repeating ourselves.

betesh-at-data-axle:
If you look at the build output, this is the command that starts the containers (I added some whitespace for readability--there you'll find this all on one line):

```
/usr/bin/docker create --name 9570b88642a645dd86385c1a4bef51ee_postgres_380cc6 \
    --label 8a33c1
    --network github_network_b05ff9d8690c43fd9ff62d88c1d8f886
    --network-alias postgres \
    -p 5432:5432 \
    --health-cmd pg_isready \
    --health-interval 10s \
    --health-timeout 5s \
    --health-retries 5 \
    -e "POSTGRES_USER=runner" \
    -e "POSTGRES_HOST_AUTH_METHOD=trust" \
    -e GITHUB_ACTIONS=true \
    -e CI=true \
    postgres
```

So these are all passed straight to docker.

Later, it checks health using:

````
/usr/bin/docker inspect \
    --format="{{if .Config.Healthcheck}}{{print .State.Health.Status}}{{end}}" \
    0a15515bd86b4b72cddfbccfb6c6afe1284d62d496f1ab4ef05e35f1a81a75d1
````

PR #133, opened 2021-05-28
Clean up some rake tasks.

ebarendt:
- Change :create to be idempotent, removing create_missing.
- update_mapping should do that, not create and deploy new indexes.

Follow up from https://github.com/data-axle/elastic_record/commit/9288de35731b57df6637de60961886612cff073f

### DIFF
--- a/lib/elastic_record/tasks/index.rake
+++ b/lib/elastic_record/tasks/index.rake
@@ -11,12 +11,16 @@ module ElasticRecord
 end
 
 namespace :index do
-  desc "Create index for CLASS or all models."
+  desc "Create index for all models that don't already have one"
   task create: :environment do
     ElasticRecord::Task.get_models.each do |model|
       index = model.elastic_index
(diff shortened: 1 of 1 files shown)


ebarendt:
@pumpkineater @betesh-at-data-axle - was anything built on top of the :create_missing rake task?

pumpkineater:
Not by me, it's only referenced in content_system bin/setup

betesh-at-data-axle:
@ebarendt In the [upgrade from Rails 5.2 to to Rails 6.0](https://railsdiff.org/5.2.6/6.0.0), bin/setup became idempotent, making bin/update obsolete.  To do this correctly, we had to add an idempotent way to create the elastic_search indeces.  See https://github.com/data-axle/content_system/pull/5627.  Since [you already updated it in content_system](https://github.com/data-axle/content_system/pull/6038), we should be all set.

PR #138, opened 2022-01-27
Support parent-child joins

betesh-at-data-axle:
[INFO-9454](https://infogroup.atlassian.net/browse/INFO-9454)

[ElasticSearch Reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html)

Note that this depends on [the latest arelastic PR](https://github.com/matthuhiggins/arelastic/pull/33).  I have requested a new release from the maintainer after which I will update the gemspec here to require the latest version.

@matthinea I was able to preserve your commit that renamed `parent` => `routing` without making any changes to that file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ When one model inherits from another, ElasticRecord makes some assumptions about
 These can all be overridden.  For instance, it might be desirable for the child documents to be in a separate index.
 
 ### Join fields
-ElasticSearch provides support for join queries by providing a mechanism for declaring a join field that specifies the parent-child relationship between different types of records on the same index.
+ElasticSearch supports declaring a join field that specifies a parent-child relationship between documents of different types in the same index ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html)).
 ElasticRecord provides a short-(but not-so-short)-cut for declaring the mapping:
 
 ```ruby
@@ -222,16 +222,16 @@ end
 ```
 
 `has_es_children` accepts an optional `name` argument, with a sane default. In the above example, it would default to `country`. The name can later be used to construct `has_parent` queries.
-The `join_field` will be used as the name of a method that gets defined both on the parent and on all children (and grandchildren), in addition to being the name of the mapping for the join field.  Be sure this doesn't conflict with the name of an existing method.  You are unlikely to need to use this name directly.
+ElasticRecord will define a getter method with the same name as the value provided to `join_field` on both the parent and all children (and grandchildren).
 
-`::ElasticRecord::Model::Joining::JoinChild.new`, too, accepts optional arguments with sane defaults:
-* `name`: In the above example, it would default to `state`. The name can later be used to construct `has_child` queries.
-* `children`: Can be another instance of `::ElasticRecord::Model::Joining::JoinChild` or an Array of them. Defaults to an empty Array.  Theoretically, an arbitrary number of layers of parent-child joins can be achieved this way.
-* `parent_id_accessor`: A method to call on the child to retrieve the ID of the parent. Can be a proc, which will be executed in the context of the child object, or the name of a method to be called on the child object.  In the above example, it would default to `country_id`.
+`::ElasticRecord::Model::Joining::JoinChild.new` optional arguments:
+* `name`: defaults to the snake case version of the value provided to `klass` (e.g. `state` in the example above). Can be used to construct `has_child` queries.
+* `children`: Another instance of `::ElasticRecord::Model::Joining::JoinChild` or an Array of instances. Defaults to an empty Array.  Theoretically, an arbitrary number of layers of parent-child joins can be achieved this way.
+* `parent_id_accessor`: Determines how the ID of the parent is retrieved. Can be a proc, which will be executed in the context of the child object, or a symbol corresponding to the name of a method defined on the child object.  In the above example, it would default to `country_id`.
 
-Note: Creating, deleting and updating mapping on the index must be handled via the Top-Level parent.  Running `rake index:create CLASS=State` has no effect.
-
-Note: The `load_from_source` configuration is not currently supported for indexes with a join field, though theoretically it could be implemented.
+Notes:
+* Creating, deleting and updating mapping on the index must be handled via the Top-Level parent.  Running `rake index:create CLASS=State` has no effect.
+* The `load_from_source` configuration is not currently supported for indices with a join field.
 
 ### Load Documents from Source
 
--- a/README.md
+++ b/README.md
@@ -202,6 +202,37 @@ When one model inherits from another, ElasticRecord makes some assumptions about
 
 These can all be overridden.  For instance, it might be desirable for the child documents to be in a separate index.
 
+### Join fields
+ElasticSearch provides support for join queries by providing a mechanism for declaring a join field that specifies the parent-child relationship between different types of records on the same index.
+ElasticRecord provides a short-(but not-so-short)-cut for declaring the mapping:
+
+```ruby
+class State
+  include ElasticRecord::Model
+end
+
+class Country
+  include ElasticRecord::Model
+
+  has_es_children(
+    join_field: 'pick_a_name_for_the_join_field',
+    children: ::ElasticRecord::Model::Joining::JoinChild.new(klass: State)
+  )
+end
+```
+
+`has_es_children` accepts an optional `name` argument, with a sane default. In the above example, it would default to `country`. The name can later be used to construct `has_parent` queries.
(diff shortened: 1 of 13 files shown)


test/elastic_record/model/joining_test.rb
```diff
@@ -0,0 +1,88 @@
...
+
+class ElasticRecord::Model::JoiningTest < MiniTest::Test
+  class Child < ActiveRecord::Base
+    include ElasticRecord::Model
+    self.elastic_index.mapping[:properties] = ::Widget.elastic_index.mapping[:properties].dup
+    self.table_name = 'widgets'
+    belongs_to :parent, foreign_key: :warehouse_id
+  end
+
+  class Parent < ActiveRecord::Base
```
matthinea, line 11: Maybe just name these classes `Mom` and `Son`, to make the tests simpler?

test/elastic_record/model/joining_test.rb
```diff
@@ -0,0 +1,88 @@
...
+    assert_equal [], Parent.filter(Parent.arelastic.queries.term('arbitrary', 'son')).to_a
+    assert_equal [], Child.filter(Child.arelastic.queries.term('arbitrary', 'mom')).to_a
+
+    assert_equal [parent.name], Parent.filter(Parent.arelastic.queries.match_all.has_child('son')).to_a.map(&:name)
+    assert_equal [], Parent.filter(Parent.arelastic.queries.match_all.has_parent('mom')).to_a.map(&:name)
+    assert_equal [child.name], Child.filter(Child.arelastic.queries.match_all.has_parent('mom')).to_a.map(&:name)
+    assert_equal [], Child.filter(Child.arelastic.queries.match_all.has_child('son')).to_a.map(&:name)
+
+    assert_equal [parent.name], Parent.filter(Parent.arelastic.queries.match_all).to_a.map(&:name)
+    assert_equal [child.name], Child.filter(Child.arelastic.queries.match_all).to_a.map(&:name)
```
matthinea, line 86: These are all great; we should probably add a test that verifies that the routing changes in `Index::Document` still work too https://github.com/data-axle/elastic_record/blob/16a33675a62938362cb34f62adec2868a835d457/test/elastic_record/index/documents_test.rb#L96-L95

README.md
```diff
@@ -202,6 +202,37 @@ When one model inherits from another, ElasticRecord makes some assumptions about
...
+
+`::ElasticRecord::Model::Joining::JoinChild.new`, too, accepts optional arguments with sane defaults:
+* `name`: In the above example, it would default to `state`. The name can later be used to construct `has_child` queries.
+* `children`: Can be another instance of `::ElasticRecord::Model::Joining::JoinChild` or an Array of them. Defaults to an empty Array.  Theoretically, an arbitrary number of layers of parent-child joins can be achieved this way.
+* `parent_id_accessor`: A method to call on the child to retrieve the ID of the parent. Can be a proc, which will be executed in the context of the child object, or the name of a method to be called on the child object.  In the above example, it would default to `country_id`.
+
+Note: Creating, deleting and updating mapping on the index must be handled via the Top-Level parent.  Running `rake index:create CLASS=State` has no effect.
+
+Note: The `load_from_source` configuration is not currently supported for indexes with a join field, though theoretically it could be implemented.
```
matthinea, line 234:
```suggestion
`::ElasticRecord::Model::Joining::JoinChild.new` optional arguments:
* `name`: defaults to the snake case version of the value provided to `klass` (e.g. `state` in the example above). Can be used to construct `has_child` queries.
* `children`: Another instance of `::ElasticRecord::Model::Joining::JoinChild` or an Array of instances. Defaults to an empty Array.  Theoretically, an arbitrary number of layers of parent-child joins can be achieved this way.
* `parent_id_accessor`: Determines how the ID of the parent is retrieved. Can be a proc, which will be executed in the context of the child object, or a symbol corresponding to the name of a method defined on the child object.  In the above example, it would default to `country_id`.

Notes:
* Creating, deleting and updating mapping on the index must be handled via the Top-Level parent.  Running `rake index:create CLASS=State` has no effect.
* The `load_from_source` configuration is not yet supported for indices with a join field.
```

matthinea (commented):
Good stuff! Super comprehensive. I left some feedback

lib/elastic_record/model/joining.rb
```diff
@@ -0,0 +1,108 @@
...
+              raise "parent_id_accessor must be callable for #{klass}"
+            end
+          end
+
+          name ||= klass.to_s.underscore
+
+          children = Array.wrap(children)
+          children.each do |child|
+            unless child.is_a?(self.class)
+              raise "JoinChild#children can only contain #{self.class}. Cannot be #{child}!"
```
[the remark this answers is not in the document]
betesh-at-data-axle: The current name is appropriate.  `children` are the children of the object that exposes them via the `children` method.  I would expect a `grandchildren` method of a `child` to return great-grandchildren, not grandchild, of the root object.

lib/elastic_record/relation.rb
```diff
@@ -52,7 +52,13 @@ def to_a
 
     def find_hits(search_hits)
       if klass.elastic_index.load_from_source
+        if klass.respond_to?(:es_join_field)
+          raise "Loading parent/child joins from source is not currently supported!"
```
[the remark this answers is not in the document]
betesh-at-data-axle: `yet` implies a commitment to support it in the future.  `currently` implies that it's a possibility but makes no promises.

lib/elastic_record/relation.rb
```diff
@@ -52,7 +52,13 @@ def to_a
...
+        end
         search_hits.hits.map { |hit| klass.from_search_hit(hit) }
+      elsif klass.respond_to?(:es_join_field)
+        ids = search_hits.hits.filter_map { |hit| hit['_id'] if hit.dig('_source', klass.es_join_field, 'name') == klass.es_join_name }
```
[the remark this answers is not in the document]
betesh-at-data-axle: I considered that possibility for a while.  Eventually, I decided that the Single Responsibility of a `SearchHits` object is to de-serialize the response received from ES, not to worry about which parts of the response are relevant.  It's the job of `ElasticRecord::Model`, which is concerned with behaving like an `ActiveRecord` object, to ensure that we don't call `find` on one class and get instances of another class.

lib/elastic_record/model/joining.rb
```diff
@@ -0,0 +1,108 @@
...
+            else
+              parent_id_accessor.respond_to?(:call) ? instance_exec(&parent_id_accessor) : send(parent_id_accessor)
+            end
+          end
+
+          klass.elastic_index.alias_name = parent.es_root.elastic_index.alias_name
+          klass.elastic_index.disable_index_creation = true
+          parent.es_root.elastic_index.mapping.merge!(klass.elastic_index.mapping)
+
+          children.each { |child| assign_to_parent!(parent: klass) }
```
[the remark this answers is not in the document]
betesh-at-data-axle: Good catch!  I'm a bit surprised this didn't break existing tests.  We would have caught it when writing tests that involved grandchildren, but I deliberately didn't spend time writing tests for that because it's not a feature we anticipate using in the foreseeable future.

README.md
betesh-at-data-axle: I'm ok with this except for changing `currently` to `yet`.

test/elastic_record/model/joining_test.rb
betesh-at-data-axle: Fixed.  It caught a bug--I had to add `demodulize` when inferring the relation name from the klass.

test/elastic_record/model/joining_test.rb
betesh-at-data-axle: I added a test for bulk batches.

test/elastic_record/model/joining_test.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    son = Son.new(id: 10, name: 'King Edward VII', mother: mother)
+    Son.insert_all([son.attributes])
+
+    index = Mother.elastic_index
+    assert_nil index.current_bulk_batch
+
+    index.bulk do
+      index.index_record(mother)
+      index.index_record(son)
```
betesh-at-data-axle, line 101: Note that this test is contrived to illustrate that routing works correctly for bulk batches.  In practice, this isn't quite correct usage since we're calling `Mother.elastic_index.index_record(son)`, when, in real life, we'd want to use `Son.elastic_index.index_record(son)` if this were outside of a batch.  `Mother.elastic_index` and `Son.elastic_index` are different Ruby objects, even though they reference the same ES index.

matthinea:
Bumped arelastic, now re-running github actions

matthinea (approved):

PR #142, opened 2022-03-03
Support simpler `has_es_children` input

matthinea:
### Problem

The interface to `has_es_children` is a little cumbersome. Most users won't require complex nestings and will only want to specify a single child or two.

### Solution

Support passing an instance of `Class` as the `children` argument to `has_es_children`. `README` has examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,12 +211,19 @@ class State
   include ElasticRecord::Model
 end
 
+class City
+  include ElasticRecord::Model
+end
+
 class Country
   include ElasticRecord::Model
 
   has_es_children(
-    children:   State,
-    join_field: 'pick_a_name_for_the_join_field'
+    join_field: 'pick_a_name_for_the_join_field',
+    children:   [
+      State,
+      ElasticRecord::Model::Joining::JoinChild.new(klass: City, parent_id_accessor: :country_code)
+    ]
   )
 end
 ```
--- a/README.md
+++ b/README.md
@@ -215,8 +215,8 @@ class Country
   include ElasticRecord::Model
 
   has_es_children(
-    join_field: 'pick_a_name_for_the_join_field',
-    children: ::ElasticRecord::Model::Joining::JoinChild.new(klass: State)
+    children:   State,
+    join_field: 'pick_a_name_for_the_join_field'
   )
 end
 ```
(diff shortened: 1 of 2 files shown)


lib/elastic_record/model/joining.rb
```diff
@@ -4,8 +4,19 @@ module Joining
...
 
+        def self.build_children(user_input)
+          case user_input
+          when Class, Array
+            Array.wrap(user_input).map { |child| new(klass: child) }
+          when ElasticRecord::Model::Joining::JoinChild
+            Array.wrap(user_input)
+          else
+            raise "Expected a Class, Array, or ElasticRecord::Model::Joining::JoinChild instance as first arg to has_es_children. Received #{user_input} instead."
+          end
```
betesh-at-data-axle, line 15: Why stop there?  It can be an Array where some elements are `JoinChild`'s and some are `Class`

Also, there's no need to raise here, because line 20 (the old line 9) will give the right error message.

```suggestion
          Array.wrap(user_input).map do |join_child|
            join_child.is_a?(self) ? join_child : new(klass: join_child)
          end
```

README.md
```diff
@@ -215,16 +215,18 @@ class Country
...
+    join_field: 'pick_a_name_for_the_join_field'
   )
 end
 ```
 
 `has_es_children` accepts an optional `name` argument, with a sane default. In the above example, it would default to `country`. The name can later be used to construct `has_parent` queries.
 ElasticRecord will define a getter method with the same name as the value provided to `join_field` on both the parent and all children (and grandchildren).
 
-`::ElasticRecord::Model::Joining::JoinChild.new` optional arguments:
+The `children` argument expects a Class, Array of Classes, or, for complex nestings, an instance of `::ElasticRecord::Model::Joining::JoinChild.new`.
```
betesh-at-data-axle, line 227:
```suggestion
The `children` argument expects a Class or, for complex nestings, an instance of `::ElasticRecord::Model::Joining::JoinChild.new`.  You can also pass an Array, each element of which is either a `Class` or a `::ElasticRecord::Model::Joining::JoinChild.new`.
```

Also consider updating the example on line 218.  I suggest an Array of 2 elements, 1 a Class and the other a JoinChild.

betesh-at-data-axle (approved):

PR #144, opened 2022-03-07
`class_for` shouldn't return an index whose creation has been disabled

matthinea:
### Problem

```ruby
ElasticRecord::Config.class_for('places')
=> ChildContact
```

### Solution

`class_for` can't return indices whose creation has been disabled.

### DIFF
--- a/lib/elastic_record/config.rb
+++ b/lib/elastic_record/config.rb
@@ -14,7 +14,9 @@ module ElasticRecord
       end
 
       def class_for(index_alias)
-        models.find { |model| model.elastic_index.alias_name == index_alias }
+        models.find do |model|
+          !model.elastic_index.disable_index_creation && model.elastic_index.alias_name == index_alias
+        end
       end
 
       def servers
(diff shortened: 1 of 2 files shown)


betesh-at-data-axle (commented):
In practice, how do we use this method?

I'm concerned that this isn't the most intuitive way to retrieve the top-level parent of all join fields.  Another way might be:

```
        models.find do |model|
          model.elastic_index.alias_name == index_alias && model.es_root == model
        end
```

That would be a more explicit way of expressing what we're trying to achieve here.

matthinea:
I got to thinking about what this should return, and I think the better solution is to just remove `ChildContact` from `model_names`, like we were [planning to do anyways](https://github.com/data-axle/content_system/blob/d34487b0c6cfe2ceb981bbc4b073efe8948ef9e7/config/initializers/elastic_record.rb#L19).

Reasoning: `class_for` should return `nil` when passed the name of an index that doesn't actually exist (e.g. `child_contacts`).

PR #145, opened 2022-03-08
Filter by join type

betesh-at-data-axle:
Until now, when querying indices that contain parent-child joins,
we were retriving records of multiple types and then filtering
down to the type we were interesteed in on the client-side after
retrieving results.

This was inefficient.  Going forward, we will filter server-side.


matthinea (approved):
Nice! This is really clean

PR #147, opened 2022-03-24
Add ability to safe load relation records

j1wilmot:
[JIRA](https://infogroup.atlassian.net/browse/INFO-9864)

**Problem:**

When ElasticRecord::Relation loads results #find_hits is called, which takes the IDs from ElasticSearch and calls `Model.find`. This results in an `ActiveRecord::RecordNotFound` error if Elastic has IDs that do not exist in Postgres.

**Solution:**

Add the ability to load the Relation's records without erroring on missing IDs. We could make this the default behavior for `find_hits`, but I'm concerned that will have larger repercussions (result order tests fail when defaulting to where instead of find), so initially this can be enabled where needed (I.E. for submissions).

### DIFF
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -67,7 +67,7 @@ module ElasticRecord
         if safe?
           klass.where(id: ids).to_a
         else
-          klass.find search_hits.to_ids
+          klass.find ids
         end
       end
     end
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -51,16 +51,20 @@ module ElasticRecord
       @records ||= find_hits(search_hits)
     end
 
-    def load_safe_hits
-      @records ||= find_hits(search_hits, safe: true)
+    def safe!
+      @safe = true
     end
 
-    def find_hits(search_hits, safe: false)
+    def safe?
+      @safe
+    end
+
+    def find_hits(search_hits)
       if klass.elastic_index.load_from_source
         search_hits.hits.map { |hit| klass.from_search_hit(hit) }
       else
         ids = search_hits.to_ids
-        if safe
+        if safe?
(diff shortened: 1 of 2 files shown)


lib/elastic_record/relation.rb
```diff
@@ -51,11 +51,20 @@ def to_a
...
+    def find_hits(search_hits, safe: false)
       if klass.elastic_index.load_from_source
         search_hits.hits.map { |hit| klass.from_search_hit(hit) }
       else
-        klass.find search_hits.to_ids
+        ids = search_hits.to_ids
+        if safe
+          klass.where(id: ids).to_a
+        else
+          klass.find search_hits.to_ids
```
betesh-at-data-axle, line 66:
```suggestion
          klass.find ids
```

lib/elastic_record/relation.rb
```diff
@@ -51,11 +51,20 @@ def to_a
       @records ||= find_hits(search_hits)
     end
 
-    def find_hits(search_hits)
+    def load_safe_hits
+      @records ||= find_hits(search_hits, safe: true)
```
betesh-at-data-axle, line 55: This method's name, as well as the context where it would typically be used, give no hint that calling the method has side effects on the object it's called on.  If I saw this method called somewhere, I would think it just returns the safe records, without changing state.

How about instead defining a method that changes the state of the class, without returning any value.  i.e.

```ruby
  def safe!
    @safe = true
  end

  def safe?
    @safe
  end

  def find_hits(search_hits)
    ...
    if safe?
      klass.where(id: ids).to_a
    else
      klass.find(ids)
    end
  end
```

j1wilmot: Nice! I like this interface a lot better.

betesh-at-data-axle (approved):

PR #149, opened 2022-03-29
Fix nested & negated queries for models with joins

matthinea:
### Problem

We're filtering by join field in `build_filter_nodes`. But calls to that method from `not` and `nested` result in negated and nested versions of that query, respectively.

### Solution

Don't add that field when negating. It'll still get added by the initial call to `filter`, as the tests show.

### DIFF
--- a/lib/elastic_record/relation/search_methods.rb
+++ b/lib/elastic_record/relation/search_methods.rb
@@ -19,7 +19,7 @@ module ElasticRecord
         end
 
         def add_filter_nodes_to_scope(filters)
-          filter_value = @scope.send(:build_filter_nodes, filters).map do |filter_node|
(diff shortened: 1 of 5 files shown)


lib/elastic_record/relation/search_methods.rb
```diff
@@ -225,10 +225,10 @@ def build_filter(filters)
           end
         end
 
-        def build_filter_nodes(filters)
+        def build_filter_nodes(filters, add_join_field_filter = true)
```
betesh-at-data-axle, line 228:
```suggestion
        def build_filter_nodes(filters, add_join_field_filter: true)
```

betesh-at-data-axle (approved):
Approved with a suggestion

PR #151, opened 2022-04-06
Define es_root everywhere

matthinea:
### Problem

We'll frequently keep having to check if a place has an ES root before asking for it. Instead, why don't we just make roots the roots of themselves?

### Solution

Make it so.

### DIFF
--- a/lib/elastic_record/model.rb
+++ b/lib/elastic_record/model.rb
@@ -35,6 +35,10 @@ module ElasticRecord
       def elastic_index=(index)
         @elastic_index = index
       end
+
+      def es_root
+        self
+      end
     end
 
     def index_to_elasticsearch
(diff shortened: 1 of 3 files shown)


betesh-at-data-axle (changes requested):
We are using the fact that this method is defined in [two](https://github.com/data-axle/content_system/blob/60dff967816d4571b3a9affac1dca3660b74d070/app/jobs/build_index_job/record_synchronizer.rb#L24) [places](https://github.com/data-axle/content_system/blob/60dff967816d4571b3a9affac1dca3660b74d070/app/jobs/build_index_job/batch_builder.rb#L13) to determine whether a class is the parent of an ES join.

If we define this method everywhere and want that code to continue working, we'd have to define `es_descendants` on all models as well and have it return an empty Array.  We could then remove the `respond_to?` check from each of those lines.

matthinea:
@betesh-at-data-axle I'm planning on patching that when I update this gem in Platform.

betesh-at-data-axle:
We should make 2 small changes in joining.rb:

1. Remove `define_method(:es_root)` from the parent
2. Use [redefine_method](https://github.com/rails/rails/blob/v6.1.4.7/activesupport/lib/active_support/core_ext/module/redefine_method.rb#L17)when defining `es_root` for the child.

betesh-at-data-axle (approved):

PR #154, opened 2022-04-27
Support deserialization of `Arelastic::Aggregations::Parent` aggregation responses

betesh-at-data-axle:


matthinea (approved):
Figured something like this would be necessary

betesh-at-data-axle:
@matthinea This needed a rewrite once it was clear to me how all the response from ES was formatted.  Please review again.

matthinea (approved):

PR #155, opened 2022-05-03
Filter ElasticRecord::Relation by parent and child Elastic::Relations

betesh-at-data-axle:


lib/elastic_record/relation/search_methods.rb
```diff
@@ -236,7 +236,14 @@ def build_filter_nodes(filters, add_join_field_filter = true)
             if filter.is_a?(Arelastic::Nodes::Node)
               nodes << filter
             elsif filter.is_a?(ElasticRecord::Relation)
-              nodes << Arelastic::Queries::HasChild.new(filter.elastic_index.mapping_type, filter.as_elastic['query'])
+              join_type = if klass.in?(Array.wrap(filter.klass.try(:es_descendants)))
+                 Arelastic::Queries::HasParent
+              elsif filter.klass.in?(Array.wrap(klass.try(:es_descendants)))
```
matthinea, line 241: I see a method, `descendent_of`, that you could use on L239 and L241, just swapping the order of `filter.klass` and `klass` in the method params.

lib/elastic_record/relation/search_methods.rb
```diff
@@ -236,7 +236,14 @@ def build_filter_nodes(filters, add_join_field_filter = true)
...
+                 Arelastic::Queries::HasChild
+              else
+                raise "You can't filter a #{filter.klass.name} ('#{filter.as_elastic}') inside a #{klass.name}"
```
matthinea, line 244: `filter.as_elastic` has the potential to be a huge object. Does it really belong in the error message?

matthinea (changes requested):

lib/elastic_record/relation/search_methods.rb
betesh-at-data-axle: Fixed

lib/elastic_record/relation/search_methods.rb
betesh-at-data-axle: Fixed

matthinea (approved):

PR #156, opened 2022-05-09
Change kwarg name to better reflect function

matthinea:
### Problem

Keyword argument name reflects developer intention rather than functional impact.

### Solution

Rename it.

### DIFF
--- a/lib/elastic_record/model/as_document.rb
+++ b/lib/elastic_record/model/as_document.rb
@@ -1,11 +1,11 @@
 module ElasticRecord
   module Model
     module AsDocument
-      def as_search_document(put_nil_values: false)
+      def as_search_document(send_nil_values: false)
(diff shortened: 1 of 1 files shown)


betesh-at-data-axle:
`put` implies (to me) that this kwarg is only relevant to updates, not creates.  Is that the case?  If so, would there be any disadvantage to making it even clearer by calling it `update_nil_values`?

matthinea:
It can be a [put or post](https://github.com/data-axle/elastic_record/blob/ffb981537d2ec0f7e39f7dadc91ff48c64abe0c4/lib/elastic_record/index/documents.rb#L38-L43). 

`update_nil_values` sounds to me like the `nil` values are getting updated (to something not-nil). How about `send_nil_values`?

betesh-at-data-axle (approved):

PR #157, opened 2022-05-13
Add interface to Elasticsearch's `search_after` API

matthinea:
https://infogroup.atlassian.net/browse/INFO-9681

### Problem

`scroll` is no longer recommended for deep pagination. `search_after` with a point-in-time is recommended instead.

### Solution

- Pull pagination methods out into their own module
- Add support for deep pagination via Elasticsearch `search_after` API


lib/elastic_record/index/pagination.rb
```diff
@@ -0,0 +1,62 @@
...
+        case e.status_code
+        when '400' then raise ElasticRecord::InvalidScrollError, e.message
+        when '404' then raise ElasticRecord::ExpiredScrollError, e.message
+        else raise e
+        end
+      end
+
+      def delete_scroll(scroll_id)
+        connection.json_delete('/_search/scroll', { scroll_id: scroll_id })
+      end
```
matthinea, line 21: All this is unchanged

lib/elastic_record/index/scroll_enumerator.rb
```diff
@@ -0,0 +1,59 @@
+module ElasticRecord
+  class Index
+    class ScrollEnumerator
```
matthinea, line 3: This too is unchanged, just pulled out into its own file.

lib/elastic_record/relation/batches.rb
```diff
@@ -29,11 +29,25 @@ def build_scroll_enumerator(options)
...
       end
+
+      private
+
+        def build_paginator(options)
+          if options.delete(:paginator) == :search_after
+            build_search_after(options)
+          else
+            build_scroll_enumerator(options)
+          end
```
matthinea, line 49: I think this is a good approach while we test `search_after` in production environments.

lib/elastic_record/index/search.rb
```diff
@@ -2,62 +2,6 @@
...
-          search_options = { size: batch_size, scroll: keep_alive }
-          elastic_query = @search.reverse_merge('sort' => '_doc')
-
-          @elastic_index.search(elastic_query, search_options)
-        end
-      end
-    end
-
     module Search
       def record_exists?(id)
```
betesh-at-data-axle, line 6: `record_exists?` and `explain` don't seem to be particularly related to the name or purpose of this module anymore.  Consider moving them into a separate module or into the class that includes this model.

lib/elastic_record/index/search_after.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+
+      def initial_search_response
+        @initial_search_response ||= search_after
+      end
+
+      private
+
+        def search_after
+          @elastic_index.search_after(
```
betesh-at-data-axle, line 65: This will create a lot of `SearchAfter` objects recursively, incurring the risk that we overflow the stack.  Could we try an implementation that doesn't use recursion so that we're only keeping 1 `SearchAfter` in memory at a time?

betesh-at-data-axle: I get it now.  There's a difference between `search_after` and `build_search_after`.

lib/elastic_record/index/scroll_enumerator.rb
betesh-at-data-axle: It would have been easier to review if the code that was just moved around without changes was in a separate PR.  Consider breaking this into 2 PR's.

lib/elastic_record/index/pagination.rb
```diff
@@ -0,0 +1,62 @@
+module ElasticRecord
+  class Index
+    module Pagination
```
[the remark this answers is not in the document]
matthinea: Done, except I left the `SearchAfter` logic in a module called `Pagination` since the `SearchAfter` class is already taking up the file called `search_after`.

betesh-at-data-axle:
Approved but in the future if there's going to be a lot of moving code around, please do that in a separate, prior PR.

lib/elastic_record/index/search_after.rb
```diff
@@ -0,0 +1,79 @@
...
+          response = search_after
+
+          if response['pit_id'] && response['pit_id'] != point_in_time_id
+            delete_pit
+          end
+        end
+
+        @point_in_time_id = response['pit_id']
+        @last_sort_values =
+          if last_hit = SearchHits.from_response(response).hits.last
```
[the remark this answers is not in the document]
matthinea: That won't happen because if there are any results at all, `last_sort_values` will not be `nil`, and if there are no results, the `while` loop condition will not be met.

It's a valid concern, though, so I'll add a test for it in a follow-up PR.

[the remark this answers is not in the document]
matthinea: https://github.com/data-axle/elastic_record/pull/161/files#diff-fdd3ff6d262a421e3f8583142525bb38b78129a42d1ba234cc0088899f712bf9R60-R73

PR #160, opened 2022-05-18
Safer way to disable deferring during testing

matthinea:
### Problem

We have an established, not-polluting way of disabling deferring for a test, but we're not using it everywhere.

### Solution

Move the helper method to `test_helper.rb` and use it everywhere.


betesh-at-data-axle (approved):
Do you think this will also fix the intermittent failures on CI?

matthinea:
I've only seen that one intermittent failure where the `Mother` query doesn't have the `type_of_document`, and this won't impact that.

PR #163, opened 2022-07-05
Fix deletion with routing

matthinea:
### Problems

- `delete_by_query` doesn't support deleting documents that require `routing` to be specified.
- `delete_document` improperly specifies query params when a `routing` value is specified.

### Solution

Fix 'em.

### DIFF
--- a/test/elastic_record/index/documents_test.rb
+++ b/test/elastic_record/index/documents_test.rb
@@ -60,7 +60,7 @@ class ElasticRecord::Index::DocumentsTest < MiniTest::Test
     index.index_document('abc', { color: 'red' })
     assert index.record_exists?('abc')
 
-    index.delete_document('abc')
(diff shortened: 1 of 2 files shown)


test/elastic_record/index/documents_test.rb
```diff
@@ -60,7 +60,7 @@ def test_delete_document
     index.index_document('abc', { color: 'red' })
     assert index.record_exists?('abc')
 
-    index.delete_document('abc')
+    index.delete_document('abc', routing: 'abc')
     refute index.record_exists?('abc')
```
matthinea, line 64: Confirmed that this change to L63 causes the assertion on L64 to fail on master.

j1wilmot (approved):

betesh-at-data-axle (approved):

PR #164, opened 2022-07-12
preserve order of records when doing safe query

aschlick:
[JIRA](https://infogroup.atlassian.net/browse/INFO-10493)

### Problem

Making a relation `safe` removes any ordering within the query, instead orders based on database internal ordering. 

### Solution

Resort the results from the database based on that from Elastic. 

This is demonstrated in data-axle/content_system#6982

### DIFF
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -70,7 +70,7 @@ module ElasticRecord
       else
         ids = search_hits.to_ids
         if safe?
-          klass.where(id: ids).to_a
+          klass.where(id: ids).to_a.sort_by { |doc| ids.index { |i| i.to_s == doc.id.to_s } }
         else
           klass.find ids
         end
--- a/test/elastic_record/relation_test.rb
+++ b/test/elastic_record/relation_test.rb
@@ -42,13 +42,16 @@ class ElasticRecord::RelationTest < MiniTest::Test
   end
 
   def test_to_a
-    Widget.create!(color: 'red')
+    Widget.create!(color: 'green')
     Widget.create!(color: 'red')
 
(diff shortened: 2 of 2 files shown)


lib/elastic_record/relation.rb
```diff
@@ -70,7 +70,7 @@ def find_hits(search_hits)
       else
         ids = search_hits.to_ids
         if safe?
-          klass.where(id: ids).to_a
+          klass.where(id: ids).to_a.sort_by { |doc| ids.index { |i| i.to_s == doc.id.to_s } }
```
betesh-at-data-axle, line 73: Should we do something similar on line 75?

aschlick: It didn't appear that we needed to. Here is the testing I did in console around it:
```
[6] pry(#<ElasticRecord::RelationTest>)> Widget.find Widget.elastic_search.to_ids
=> [#<Widget:0x0000000107bd5350 id: 2006, warehouse_id: nil, name: nil, color: "green", price: nil, widget_part: nil>,
 #<Widget:0x0000000107bd51e8 id: 2007, warehouse_id: nil, name: nil, color: "red", price: nil, widget_part: nil>]
[7] pry(#<ElasticRecord::RelationTest>)> Widget.find Widget.elastic_search.order(color: :desc).to_ids
=> [#<Widget:0x0000000107cdff98 id: 2007, warehouse_id: nil, name: nil, color: "red", price: nil, widget_part: nil>,
 #<Widget:0x0000000107ce4098 id: 2006, warehouse_id: nil, name: nil, color: "green", price: nil, widget_part: nil>]
```
I added to the `to_a` test as well to make sure that's true.

betesh-at-data-axle (approved):

PR #166, opened 2023-01-07
Format `ElasticRecord::ConnectionError` messages as JSON

matthinea:
### Problem

When an `ElasticRecord::ConnectionError` gets raised, its error message is one of two things:
* a string representation of a Ruby hash
* a string representation of a Ruby hash, with some additional non-hash text appended to it

This is inconvenient if a developer wants to dig around in the error message, which is a likely scenario, seeing as this error covers lots of Elasticsearch query shard execptions and parse exceptions.

### Solution

* Raise `ConnectionError` in a consistent way
* Pass the raw JSON to the error as its message
* Concatenate the payload (when present) to the error message such that it's still valid JSON


betesh-at-data-axle (changes requested):
Should we make this a breaking change by requiring `payload` in ConnectionError's initializer?  That would force clients that are raising it directly (if any are) to also include the payload in the error message.

matthinea:
@betesh-at-data-axle sometimes there isn't a payload, which would force users to pass `nil`. So, I don't think it's really needed

lib/elastic_record/errors.rb
```diff
@@ -4,11 +4,27 @@ class Error < StandardError
...
+      def build_message_hash(json_error, json_payload)
+        error   = JSON.parse(json_error)
+        payload = JSON.parse(json_payload || '{}')
+
+        error.merge!('payload' => payload)
+      rescue JSON::ParserError
+        {
+          'error'   => json_error,
+          'payload' => json_payload
+        }
```
betesh-at-data-axle, line 26: If one is parse-able and the other isn't, no point in not parsing the one that we can parse:

```suggestion
        payload = if json_payload
          begin
            JSON.parse(json_payload)
          rescue JSON::ParserError
            json_payload
          end
        else
          nil
        end
        
        begin
          JSON.parse(json_error).merge(payload:)
        rescue JSON::ParserError
          { error: json_error, payload: }
        end
```

lib/elastic_record/errors.rb
```diff
@@ -4,11 +4,29 @@ class Error < StandardError
...
+      message = build_message(json_error, json_payload)
+
       super(message)
     end
+
+    private
+
+      def build_message(json_error, json_payload)
+        error   = { elasticsearch_response: parse_or_return(json_error) }
+        payload = { request_payload:        parse_or_return(json_payload) }
```
matthinea, line 20: If payload isn't provided, this will just return `nil`; so, no need to set a default.

betesh-at-data-axle (approved):
Approved with one request

PR #169, opened 2023-03-03
after_commit all the things

andrek-data-axle:
### Problem

`after_save` and its friends may be called even after the transaction encapsulating the database operation is rolled back.

### Solution

Use `after_commit`

### Notes

Do not use after_create_commit etc because https://www.kostolansky.sk/posts/unexpected-after-commit-behaviour/

### Links

https://api.rubyonrails.org/v6.0.6.1/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_commit
https://jakeyesbeck.com/page10/#aftercommit
https://blog1.westagilelabs.com/when-to-use-after-commit-in-rails-f5e53a22bb9

### DIFF
--- a/lib/elastic_record/model/callbacks.rb
+++ b/lib/elastic_record/model/callbacks.rb
@@ -2,12 +2,12 @@ module ElasticRecord
   module Model
     module Callbacks
       def self.included(base)
-        return unless base.respond_to?(:after_save) &&  base.respond_to?(:after_destroy)
+        return unless base.respond_to?(:after_commit)
 
         base.class_eval do
-          after_create :index_to_elasticsearch
-          after_update :update_index_document, if: -> { respond_to?(:saved_changes?) ? saved_changes? : changed? }
-          after_destroy :delete_index_document
+          after_commit :index_to_elasticsearch, on: :create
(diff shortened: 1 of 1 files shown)


j1wilmot:
Yes, makes sense to use after_commit when trying to sync multiple systems. I'm curious what performance impact of this could be.

lib/elastic_record/model/callbacks.rb
```diff
@@ -2,12 +2,12 @@ module ElasticRecord
   module Model
     module Callbacks
       def self.included(base)
-        return unless base.respond_to?(:after_save) &&  base.respond_to?(:after_destroy)
+        return unless base.respond_to?(:after_commit)
```
betesh-at-data-axle, line 5: elastic_record only supports Ruby 2.7+: https://github.com/data-axle/elastic_record/blob/3a2de444e2d6e3781f65f09611d30f5ba06d24a9/elastic_record.gemspec#L9

Based on https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html, that pretty much limits it to Rail >= 6.

So if the point of this line is to ensure that we're using a version of Rails that supports `after_commit` (which has been around since Rails 3), we can remove it.

Otherwise, this is the same as `if defined?(ActiveRecord::Base) && base < ActiveRecord::Base`, which I think is a more explicit way of expressing what we're trying to achieve here.

PR #170, opened 2023-03-03
Use ActiveSupport to encode JSON

j1wilmot:
**Problem:**
`ActiveSupport::TimeWithZone` objects do not get serialized in ISO8601 format when using JSON.generate. Use of Oj has hidden this deficiency previously.

**Solution:**
Use ActiveSupport JSON encoding when generation data for elastic search requests.

### DIFF
--- a/elastic_record.gemspec
+++ b/elastic_record.gemspec
@@ -20,4 +20,5 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'arelastic', '>= 3.4.0'
   s.add_dependency 'activemodel'
+  s.add_dependency 'activesupport'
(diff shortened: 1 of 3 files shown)


elastic_record.gemspec
```diff
@@ -20,4 +20,5 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'arelastic', '>= 3.4.0'
   s.add_dependency 'activemodel'
+  s.add_dependency 'activesupport'
```
betesh-at-data-axle, line 23: With this change, we can now change the line above it to add_development_dependency.

PR #172, opened 2023-03-10
Return block's return value from Index#bulk

andrek-data-axle:
Right now,`Widget.elastic_index.bulk do` block returns 'nil`. If I want to return a value from within it, I have to do something like this:

```
saved_widgets = [].tap do |saved_widgets|
  Widget.elastic_index.bulk do
    saved_widgets.concat(widget.filter(&:save))
  end
end
```

The proposed change fixes `ElasticRecord::Index#start_new_bulk_batch` to return the value returned from the `yield`ed block making the following code possible:

```
saved_widgets = Widget.elastic_index.bulk do
  widget.filter(&:save)
end
```

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -114,12 +114,12 @@ module ElasticRecord
         def start_new_bulk_batch(options, &block)
           connection.bulk_actions = []
 
-          yield
-
-          if current_bulk_batch.any?
-            body = current_bulk_batch.map { |action| "#{ActiveSupport::JSON.encode(action)}\n" }.join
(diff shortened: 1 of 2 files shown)


j1wilmot (approved):
Seems like a nice improvement. The current return of `verify_bulk_requests` won't be meaningful. Only concern would be if this could be misused somehow, but I think your example is a good one of how this has benefits.

It looks like [ActiveRecord::Base.transaction](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-transaction) returns the result of the block, so seems reasonable to make #bulk consistent.

matthinea (approved):

betesh-at-data-axle (approved):

PR #175, opened 2023-04-02
Clone shared search values with *VALUE_METHODS

bullfight:
### Problem

On #147 @j1wilmot added support for `safe` and #153 @matthinea added metamorphose to copy the value of `@safe` to a newly created Relation. Relation already has support for this as part of `*_value` methods defined with `SINGLE_VALUE_METHODS`. These methods are set on the `values` hash and copied to the new object when an active record method is called.

### Solution

* Defined `:safe` on `SINGLE_VALUE_METHODS`
* Move definition of `safe!` and `safe?` to `SearchMethods`
* Clone relations with klass.new

P.S. Howdy guys! I was reading through the repo and saw this little bike shed.

### DIFF
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -31,6 +31,10 @@ module ElasticRecord
       became
     end
 
+    def metamorphose(new_klass)
+      self.class.new(new_klass, values: values.dup)
+    end
+
     def count
       search_hits.total
     end
--- a/lib/elastic_record/relation.rb
+++ b/lib/elastic_record/relation.rb
@@ -31,12 +31,6 @@ module ElasticRecord
       became
     end
 
-    def metamorphose(new_klass)
-      result = self.class.new(new_klass, values: values.dup)
-      result.safe! if safe?
-      result
-    end
-
     def count
       search_hits.total
     end
@@ -58,14 +52,6 @@ module ElasticRecord
 
     def to_a
       @records ||= find_hits(search_hits)
-    end
-
-    def safe!
-      @safe = true
-    end
-
-    def safe?
-      @safe
     end
 
     def find_hits(search_hits)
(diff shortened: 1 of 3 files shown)


lib/elastic_record/relation.rb
```diff
@@ -31,12 +31,6 @@ def becomes(klass)
       became
     end
 
-    def metamorphose(new_klass)
-      result = self.class.new(new_klass, values: values.dup)
-      result.safe! if safe?
-      result
-    end
```
matthinea, line 38: I think a `metamorphose` method is still helpful, even if it's now just doing L35.

matthinea (commented):
Nice! Thanks 😄

lib/elastic_record/relation.rb
bullfight: What do you think about overriding `initialize_dup`? [Article here](https://blog.appsignal.com/2019/02/26/diving-into-dup-and-clone.html).

matthinea: How would that look? Right now we're doing something like:

```ruby
elastic_relation.metamorphose(Contact)
```

It's nice to be able to pass the class it's morphing into.

bullfight: oh, give `ElasticRecord::Relation#becomes` a try.

https://github.com/data-axle/elastic_record/blob/b7ea102f288ff1e35a346c1c904932515a95e8b3/lib/elastic_record/relation.rb#L27-L32

matthinea: Yeah, it's the `@klass` value that we want to set. (Which confusingly here is different from the `klass` value)

bullfight: Just coming back to this after a bit, I added `Relation#metamorphose` back.

matthinea (approved):
This is great thank you 😃

PR #177, opened 2024-01-22
Order `search_after` by shard doc

matthinea:
## Problem

`search_after` will only iterate over a fraction of the documents in its scope if it lingers on any given page for longer than 0.05 seconds or so.

Actual counted objects will be much lower than the value in `some_scope.count`:

```ruby
count = 0

some_scope.find_each(paginator: :search_after, batch_size: 500, keep_alive: '2m') do
  count += 1
  sleep(0.05)
end
```

## Solution

Order by `_shard_doc` when paginating using `search_after`.

```ruby
count = 0

some_scope.order('_shard_doc').find_each(paginator: :search_after, batch_size: 500, keep_alive: '2m') do
  count += 1
  sleep(0.05)
end
```

According to the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after), we shouldn't need to supply this explicitly. But, it seems to fix the issue, and it should also provide an optimization:
Screenshot 2024-01-22 at 11 18 41 AM


matthinea:
The test suite is currently broken, but I fixed it in https://github.com/data-axle/elastic_record/pull/178

That aside, I'm going to spend a few minutes seeing if I can reproduce this in a test that doesn't take longer than a minute or so. So far I've only reproduced it in a production setting iterating over 10,000 documents, and I imagine the production shard refreshes are a factor in documents not showing up in pagination. We'll see.

matthinea:
I can't reproduce the bug this PR fixes in a unit test. Even with 10_000 documents and occasional index refreshes during pagination, the issue doesn't arise. However, you can see it in a Content System console using any scope with a few thousand or more records.

j1wilmot (approved):

PR #182, opened 2025-07-04
Upgrade CI to Ubuntu 22.04

matthinea:
20.04 is no longer supported https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'dotenv-rails'
 gem 'arelastic'
 gem 'oj'
 gem 'pg'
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -6,6 +6,7 @@ gem 'dotenv-rails'
 gem 'arelastic'
 gem 'oj'
 gem 'pg'
+gem 'concurrent-ruby', '1.3.4'
```
matthinea, line 9: [See](https://github.com/rails/rails/issues/54272)

PR #184, opened 2026-06-04
Send `_bulk` actions in batches

matthinea:
## Problem

It's deceptively easy to load hundreds of thousands of records into a single `_bulk` request. These can result in `413: Request too large exception`, failing silently.

## Solution

Batch `_bulk` requests and only send max 1000 records at a time.

### DIFF
--- a/lib/elastic_record/index/documents.rb
+++ b/lib/elastic_record/index/documents.rb
@@ -121,16 +121,18 @@ module ElasticRecord
           connection.bulk_actions = nil
         end
 
+        ACTIONS_PER_BULK = 1_000
+
         def json_post_bulk(options)
-          # chunk into 1000 actions at a time
-          # add error handling for "client intended to send too large body"
-          body    = current_bulk_batch.map { |action| "#{ActiveSupport::JSON.encode(action)}\n" }.join
-          results = connection.json_post("/_bulk?#{options.to_query}", body)
+          current_bulk_batch.each_slice(ACTIONS_PER_BULK) do |actions|
+            body    = actions.map { |action| "#{ActiveSupport::JSON.encode(action)}\n" }.join
+            results = connection.json_post("/_bulk?#{options.to_query}", body)
 
-          if results.is_a?(Hash)
-            errors = results['items'].select { |item| item.values.first['error'] }
+            if results.is_a?(Hash)
+              errors = results['items'].select { |item| item.values.first['error'] }
 
-            raise ElasticRecord::BulkError.new(errors) unless errors.empty?
+              raise ElasticRecord::BulkError.new(errors) unless errors.empty?
+            end
           end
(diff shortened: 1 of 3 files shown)


lib/elastic_record/index/documents.rb
```diff
@@ -68,7 +66,7 @@ def delete_document(id, routing: nil, index_name: alias_name)
         if batch = current_bulk_batch
           instructions = { _index: index_name, _id: id, retry_on_conflict: 3 }
           instructions[:routing] = routing if routing
-          batch << { delete: instructions }
+          batch << [{ delete: instructions }]
```
matthinea, line 69: `index_document` and `update_document` bulk actions require 2 lines each, but `delete` bulk actions only require 1. By grouping them together as they're added, we can paginate them correctly at HTTP request time

test/elastic_record/index/documents_test.rb
```diff
@@ -100,6 +106,42 @@ def test_bulk
...
+        index.update_document '5', { color: 'blue' }
+        index.update_document '5', { color: 'orange' }
+        index.update_document '4', { color: 'black' }
+      end
+
+      assert_equal 3, post_calls.length
+      expected_body = <<~DOC
+        {"index":{"_index":"widgets","_id":"5"}}
+        {"color":"green"}
+        {"delete":{"_index":"widgets","_id":"3","retry_on_conflict":3}}
```
matthinea, line 133: Here we can see that the single-action `delete` bulk is correctly grouped with the 2-line `index` action before it.

lib/elastic_record/index/documents.rb
```diff
@@ -121,14 +119,20 @@ def start_new_bulk_batch(options, &block)
...
 
+        ACTIONS_PER_BULK = 1_000
+
         def json_post_bulk(options)
-          body    = current_bulk_batch.map { |action| "#{ActiveSupport::JSON.encode(action)}\n" }.join
-          results = connection.json_post("/_bulk?#{options.to_query}", body)
+          slice_size = options.delete(:actions_per_bulk) || ACTIONS_PER_BULK
+
+          current_bulk_batch.each_slice(slice_size) do |actions|
+            body    = actions.flatten.map { |action| "#{ActiveSupport::JSON.encode(action)}\n" }.join
```
j1wilmot, line 128: Would be nice to extract bulk post to a separate method so that slice behavior can be done optionally. Then method with slicing calls extracted method.

j1wilmot (approved):
LGTM!

lib/elastic_record/index/documents.rb
matthinea: I think we'll want to always slice though, no? In any case, `actions_per_bulk` can be specified if different slice sizes are needed

PR #185, opened 2026-06-04
Raise `ConnectionError` on most non-200 errors

matthinea:
## Problem

Most methods interface with Elasticsearch via `json_request`. The `json_request` method only raises on 5xx errors, or when there's an explicity `error` key in the JSON response. This leaves 3xx and 4xx errors as silent failures.

## Solution

* For `post` and `put` methods, raise unless the response code is 2xx.
* For 'get' and 'delete' methods, additionally allow 404 Not Found, since apparently some methods rely on that.

### DIFF
--- a/lib/elastic_record/connection.rb
+++ b/lib/elastic_record/connection.rb
@@ -22,7 +22,7 @@ module ElasticRecord
     end
 
     def json_get(path, json = nil)
-      json_request :get, path, json
+      json_request :get, path, json, consider_ok: 404
     end
 
     def json_post(path, json = nil)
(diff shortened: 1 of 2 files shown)


lib/elastic_record/connection.rb
```diff
@@ -42,6 +42,8 @@ def json_request(method, path, payload)
 
       response = http_request_with_retry(method, path, payload)
 
+      raise_connection_error(response, payload) unless (200..299).cover?(response.code.to_i)
```
matthinea: Unfortunately, there are a few valid use cases for non-500 response codes, like the below that goes through `head`:

https://github.com/data-axle/elastic_record/blob/ffb981537d2ec0f7e39f7dadc91ff48c64abe0c4/lib/elastic_record/index/manage.rb#L37-L40

By putting the non-200 handling in `json_request`, we get a middle-of-the-road approach that allows `http_request_with_retry` to continue to return 3xx and 4xx HTTP codes without raising, so that it can continue to be used in `head` requests as needed.

j1wilmot (commented):
Approach looks good, but should this be grouped with the 5xx handling? Now there are multiple code locations that handle status codes.

matthinea:
In principal, yes, but core issue is:

* 404 is expected to be valid for `get` and `delete`
* _all_ responses should be valid for `head`
* only 2xx responses should be valid or `put` or `post`

Let me take another pass at this and see if it can be simplified without resulting in major changes to the interface.

matthinea:
Ok it looks like we can get by with just 404 being valid for `head`, which simplifies things. Let's me test this branch on QA in a downstream application and make sure everything's working.

matthinea:
Ok, I added some test cases with real-life examples of how & why GET and DELETE should sometimes not error on 404. 

I ran some bulk and individual updates on QA documents with this branch and everything seems to be performing good 👍

PR #186, opened 2026-09-22
Fix thread safety bug in Bulk Actions

matthinea:
### Problem

ElasticRecord stores bulk batches on a connection that's shared between threads. Multi-threaded processes with concurrent bulk batches risk mixing, flushing, and discarding other threads' actions.

### Solution

Store bulk actions in thread-local, connection-scoped storage, preserving connection isolation while preventing concurrent workers from sharing a batch.


test/elastic_record/connection_test.rb
```diff
@@ -11,6 +11,30 @@ def test_options
...
+    connection.bulk_actions = [:main]
+
+    worker_actions = Thread.new do
+      connection.bulk_actions = [:worker]
+      connection.bulk_actions
+    end.value
+
+    assert_equal [:worker], worker_actions
+    assert_equal [:main], connection.bulk_actions
+  end
```
matthinea, line 25: I confirmed that this test fails on `master`

test/elastic_record/index/documents_test.rb
```diff
@@ -214,6 +214,32 @@ def test_bulk_return_value
...
+      thread.join
+
+      main_ids = index.current_bulk_batch.map { |action| action.dig(0, :index, :_id) }
+      assert_equal ['main'], main_ids
+    end
+
+    assert_empty errors
+    assert_equal ['other'], other_ids.pop
+    assert_nil index.current_bulk_batch
+  end
```
matthinea, line 241: This test also fails on `master` branch

